PR #1, opened 2016-07-12
Add analytics

kwalrath:
We might want to change to share a tracking ID with www, but for now, this is a simple solution that should work.

@tbeck & @filiph, same questions apply here as to https://github.com/dart-lang/site-www/pull/1.

Also, I noticed that this code (which is an exact copy of what the Analytics site tells me to use) uses https:// instead of //. For consistency's sake, I might change the www code to match.


kwalrath:
Yikes, I see that travis-ci/push is running!

We shouldn't push anything on non-master branches. In fact, we might consider creating a magic, non-master branch for pushes.

kwalrath:
I'm going to commit this. We can always improve it later.

PR #62, opened 2016-11-14
support for full site (qs, guide, tutorial) via angularJS

chalin:
Best to review individual commits. Update to jquery was meant to be temporary, to see if it could fix a page rendering problem I had, but the problem turned out to be due to something else. I've kept the jquery refresh since it doesn't appear to hurt.

This PR contains autosync'd .jade files from angular.io.
Use `gulp _get-resources` to get resources from your local angular.io repo.


src/_includes/head.html
```diff
@@ -43,10 +43,49 @@
...
+  -->
+  <script src="/resources/js/vendor/lodash.js"></script>
+  <script src="/resources/js/vendor/clipboard.min.js"></script>
+  <!-- Angular Material Dependencies -->
+  <script src="https://ajax.googleapis.com/ajax/libs/angularjs/1.4.8/angular.min.js"></script>
+  <script src="https://ajax.googleapis.com/ajax/libs/angularjs/1.4.8/angular-animate.min.js"></script>
+  <script src="https://ajax.googleapis.com/ajax/libs/angularjs/1.4.8/angular-aria.min.js"></script>
+  <script src="https://ajax.googleapis.com/ajax/libs/angular_material/1.0.0/angular-material.min.js"></script>
+  <!-- Firebase -->
+  <script src="https://cdn.firebase.com/js/client/2.2.4/firebase.js"></script>
```
ericjim: we don't need firebase in this case, since it is only used in `/resources` which is unique to angular.io

src/_includes/head.html
```diff
@@ -43,10 +43,49 @@
...
+  <!-- AngularFire -->
+  <script src="https://cdn.firebase.com/libs/angularfire/1.2.0/angularfire.min.js"></script>
```
ericjim: Same here

PR #87, opened 2016-11-15
Fix double hyphen for analyze-only option for dart2js

koorgoo:


chalin:
@koorgoo - thanks for the fix! Could you also submit this fix to the master branch (as a separate PR)?

koorgoo:
Yes, of course. Will do later today.

PR #184, opened 2016-12-02
docs(comp-styles): adjust TOC level

chalin:


src/angular/_jade/ts/_cache/guide/component-styles.jade
```diff
@@ -11,7 +11,7 @@ block includes
   
   In this chapter we learn how to load and apply these *component styles*.
   
-  # Table Of Contents
```
kwalrath: On this site we don't need a TOC, since it's auto-generated, right?

chalin: Right. Mind if I fix that on the next round (since there are a few pages that have inlined TOCs)?

kwalrath: SGTM.

PR #185, opened 2016-12-02
smaller pngs

kevmoo:


kevmoo:
@kwalrath that removed 600K total. 😀

PR #188, opened 2016-12-02
Switch from red shield to blue shield

kwalrath:
cc @matanlurey


kevmoo:
@kwalrath Remove the old SVG? Still used?

chalin:
@kevmoo - there is nothing to remove at this point since the entire images folder of angular.io/resources is imported. When we cut the chord, we can curate then (IMHO).

PR #191, opened 2016-12-02
[ng] page front matter cleanup

chalin:
- Drop unnecessary `angular: true` property. We're using `layout == angular` instead.
- Drop obsolete JS from `_include/header.html` that was used by old QuickStart to show/hide details.
- Only include quotes when necessary in Jekyll front matter property strings.


chalin:
@kevmoo - can you merge?

kevmoo:
I'll trust you – I also have zero clue.

You'd verified and staged and such?

chalin:
Yes, staged and verified.

PR #197, opened 2016-12-06
[ng] fix layout of prev/next links and page header for narrow displays

### LINKED ISSUE #190: /angular looks bad on narrow screens
When you go to https://webdev.dartlang.org/angular (or any other 2+-line titled page under /angular), the Angular icon overlaps with the header text when the screen size is very narrow. This is noticeable on a phone in portrait mode.

image

The image is smaller (and the appearance is fine) on pages like https://webdev.dartlang.org/angular/guide/user-input that have shorter titles.

cc @kasperl, @ericjim, @chalin

chalin:
Fixes #190 

cc @kwalrath @ericjim 

Before this PR the subnav would look like this:
> screen shot 2016-12-05 at 6 39 07 pm

----

With the PR the subnav and page heading now look like this (notice the ellipses):
- screen shot 2016-12-05 at 10 43 06 pm
- screen shot 2016-12-05 at 10 42 55 pm


kwalrath:
Looks great! Thank you!

PR #200, opened 2016-12-06
Add OS tabs support

filiph:
Also, include src/_guides/get-started-new.md as the starting point for the new Getting Started experience.


src/_guides/get-started-new.md
```diff
@@ -0,0 +1,69 @@
...
+title: "Get Started"
+description: "A guide to get you quickly writing web apps in AngularDart."
+---
+
+
+
+## 1. Install the Dart SDK
+
+<ul class="tabs__top-bar">
+    <li class="tab-link current" data-tab="tab-sdk-install-windows">Windows</li>
```
chalin: Windows by default?

src/_guides/get-started-new.md
```diff
@@ -0,0 +1,69 @@
...
+
+   With [Homebrew](http://brew.sh/),
+   installing Dart is easy.
+   
+   {% prettify shell %}
+   $ brew tap dart-lang/dart
+   $ brew install dart --with-content-shell --with-dartium
+   {% endprettify %}
+   
+   Another option is [downloading Dart manually](/install/archive).
```
chalin: URL DNE. You probably want `{{site.dartlang}}/install/archive`.

src/_assets/javascripts/main.js
```diff
@@ -129,5 +129,36 @@ $(document).ready(function(){
...
+
+    $("li[data-tab='tab-sdk-install-" + osName + "']").addClass('current');
+    $('#tab-sdk-install-' + osName).addClass('current');
+  }
+
+  if (window.navigator.userAgent.indexOf("Mac") != -1) {
+    selectOperatingSystemInTabs('mac');
+  } else if (window.navigator.userAgent.indexOf("Linux") != -1) {
+    selectOperatingSystemInTabs('linux');
+  }
```
chalin: Our ready function is getting a bit long. Maybe refactor this new code into a separate function?

src/_assets/javascripts/main.js
```diff
@@ -129,5 +129,36 @@ $(document).ready(function(){
...
+
+  // The following selects the correct default tab in /guides/get-started
+  function selectOperatingSystemInTabs(osName) {
+    clearTabsCurrent();
+
+    $("li[data-tab='tab-sdk-install-" + osName + "']").addClass('current');
+    $('#tab-sdk-install-' + osName).addClass('current');
+  }
+
+  if (window.navigator.userAgent.indexOf("Mac") != -1) {
```
chalin: Which OS is most common? Why not test explicitly for other OSs and then fall through to the most common OS?

kwalrath: Last time I tested, Windows was most common, with Linux & Mac roughly tied for 2nd/3rd.

chalin: Interesting. I would not have thought. Thanks for the info.

kwalrath: (Actually, Mac leads Linux. But they're in the same general ballpark.)

src/_guides/get-started-new.md
kwalrath: This page won't be visible, so I can fix this when I incorporate it.

src/_assets/javascripts/main.js
kwalrath: @filiph do you plan to do this?

src/_guides/get-started-new.md
kwalrath: When I wrote code like this before, I recall defaulting to... Linux? Because it was easy to tell whether something was Windows or Mac, but Linux was not consistent.

But, since Linux is the least common OS, and since phone readers might be using Linux but interested in Windows, defaulting to Windows is probably the right call.

src/_assets/javascripts/main.js
filiph: I'll do it. Good suggestion!

filiph:
I just refactored the JavaScript. 

&lt;sidenote> Wading through the code (and fixing an evident bug) doesn't make me happy about the agency's work, to be honest. My dream is still to find time and reimplement the whole site client-side logic in Dart. &lt;/sidenote>

filiph:
Re: staging. I don't have a good place to stage this right now. @kwalrath will take over, though, so she can use her staging server.

chalin:
+1 for implementing this in Dart!

filiph:
@chalin 👍 Can I get a LGTM for this PR?

chalin:
@filiph - could you stage it first? You have full access to the dev server so you can:
- build the site
- choose the dev firebase project: `firebase use dev`
- `firebase deploy`

filiph:
@chalin Aha! Okay, I staged it here: https://angulardart-org-dev.firebaseapp.com/guides/get-started-new

chalin:
Thanks for the refactoring. Here are some suggestions about presentation.

(1) The entire tab content is gray so we lose the distinction between the code and the prose; e.g.
> screen shot 2016-12-08 at 6 37 07 am

Why not use something like this:
> screen shot 2016-12-08 at 6 36 50 am

I.e., set the
- tab background to `$brand-primary` and text color `$white`;
- the current tab content will have `border: 2px solid $brand-primary`.

(2) Now that the tab content background isn't all gray, we can notice the extra blank link in the code (it happens in every tab). To remove that I suspect that you'll have to put the prettify directive on the same line as the last line of code.

(3) The linux tab content would be more readable if it were presented as prose woven with code elements something like (w/ or w/o numbering of the steps):
> 1. Enable HTTPS for apt.
> ```shell
> sudo apt-get update
> sudo apt-get install apt-transport-https
> ```
> 2. ...

(4) For linux and max shell code samples, I'd remove the `$` since that makes it easier to copy-paste.

(5) For all code examples I'd remove the indentation. Again, this makes it easier when just copy-pasting commands.

chalin:
FYI, Ward asked me to create this kind of tabbed support over on the ng.io side of things, so we might have an alternate implementation soon too.

filiph:
PTAL. It's now [staged](https://angulardart-org-dev.firebaseapp.com/guides/get-started-new).

The prose itself is there mostly as a placeholder. @kwalrath already has some prose of her own. This PR is about the tabs themselves, and a 'typical' setup.

chalin:
Ok

PR #213, opened 2016-12-07
[ng] fix broken links

### LINKED ISSUE #202: Dead link in architecture section angular of guide
https://webdev.dartlang.org/angular/guide/architecture has a link to https://webdev.dartlang.org/angular/guide/animations.html (in the "Wrap Up" section at the end), which doesn't exist.

The source seems to be [src/angular/_jade/ts/_cache/guide/architecture.jade#560](https://github.com/dart-lang/site-webdev/blob/4b7cd0bd8d0257a920bed6cafcaf806e68e1ebd3/src/angular/_jade/ts/_cache/guide/architecture.jade#L560).

### LINKED ISSUE #204: [ng] address broken links in angular docs
Subsumes #202

chalin:
- fix to links in general webdev pages
- add Dart Jade page even if .hide is true

Fixes #202 
Fixes #204


src/angular/guide/style-guide.jade
```diff
@@ -0,0 +1,9 @@
+---
```
kwalrath: I don't think we should have a style-guide page.

src/angular/guide/browser-support.jade
```diff
@@ -0,0 +1,9 @@
+---
```
kwalrath: I don't think we should have this page (for now, at least). I'd just assume that AngularDart supports any browser that Dart supports.

src/angular/guide/testing.jade
```diff
@@ -0,0 +1,9 @@
+---
```
kwalrath: Our testing story is going to be different from the TS testing story, so let's not include this page for now. (We'll definitely want a testing page, but it shouldn't refer to the TS page.)

src/angular/guide/upgrade.jade
```diff
@@ -0,0 +1,9 @@
+---
```
kwalrath: Not appropriate for Dart. It's possible we (or someone) will provide some help, but it might never be in this doc set.

src/angular/guide/animations.jade
```diff
@@ -0,0 +1,9 @@
+---
```
kwalrath: The "it in Typescript" link in this page takes me to the [TS doc homepage](https://angular.io/docs/ts/latest/) instead of the [TS Animations page](https://angular.io/docs/ts/latest/guide/animations.html).

(I just checked the Security page, and it's the same.)

chalin: Fixed

src/angular/guide/testing.jade
chalin: I went with an "under construction" page.

PR #215, opened 2016-12-08
ensure site deploy is built in production mode

### LINKED ISSUE #199: [ng] Travis - don't try to deploy from ng-docs
This isn't a problem other than the potentially confusing message that a deploy attempt is skipped, when in fact the deploy for master succeeded.

### LINKED ISSUE #214: use Jekyll production mode when deploying via Travis

chalin:
Fixes #214
Also fixes #199


_config.yml
```diff
@@ -136,7 +136,7 @@ assets:
     - "*.woff"
     - "*.svg"
   #   - "bundle.css"
-  digest: true
+  # digest: true
```
kwalrath: This is a no-op because digest is true by default, right? We probably should propagate this change to site-www, as well, to avoid unnecessary diffs.

chalin: According to the [docs](https://jekyll.github.io/jekyll-assets/):
> - Disable digesting by default in development.
> - Digest by default in production

PR #246, opened 2016-12-15
Update user-input.jade

a14n:


kwalrath:
Thanks, @a14n! @chalin can we make changes like this (finally!) or do we still have tendrils of ng.io stopping that?

chalin:
No strings attached for the files under `src/angular` with the exception of the files below `src/angular/_jade` (which may require a little extra care).

Thanks for the PR Alexandre. Note that I'd go further any change:
> We cast ...

to

> We declare ...

since there is no casting involved.

a14n:
@chalin updated

a14n:
humm perhaps _We declare the `$event` **with** a `dynamic` type_ would be better ?

chalin:
Ah, right, that does need some tweaking. I'd propose:
> We declare the type of `$event` as `dynamic`, which means ...

kwalrath:
"casting" is mentioned later, too.

But actually... why are we using `dynamic`? I thought we never recommended that for user code, and used `Object` instead.

chalin:
Yes, casting is mentioned again, but I was aiming to just get this simple fix in, with the objective of making a full pass later on. But if you prefer that it be fixed in one go, that is ok with me.

It feels as if this version of the page was influenced by the TS counterpart. I think that the use of `dynamic` is ok here as it covers the scenario where we're just not bothering to think about what the type should be.

kwalrath:
Full pass later on SGTM.

PR #249, opened 2016-12-16
chore: use linkcheck under Travis; drop html_proof.rb

### LINKED ISSUE #247: Use linkcheck instead of html_proof
Supersedes #238 and #245.

Once https://github.com/filiph/linkcheck/issues/2 is addressed, we'll be able to screen out false positives and known errors too.

chalin:
Fixes #247 

cc @kwalrath @kevmoo


chalin:
This PR re-enables link checking under Travis; i.e., if there invalid links (not present in the skip file), then the build will fail.

PR #250, opened 2016-12-16
[ng] docs(template-syntax): fix filename for code excerpt

chalin:
Main fix is to file name (incorrectly) previously shown as `HeroDetailComponent.dart`.


Sfshaza:
lgtm

Issue #263, opened 2016-12-16
[ng] create security page


chalin:
_From @chalin on July 2, 2016 15:27_

- [x] proofread TS chapter - #1795
- [x] create initial version of Dart chapter prose and sample code - #1813
- [x] add sample to https://github.com/angular-examples/ and deploy

_Copied from original issue: angular/angular.io#1812_

kwalrath:
I just selected item 3 because the security example is deployed:

https://angular-examples.github.io/security/
https://github.com/angular-examples/security

kwalrath:
And the page is out:

https://webdev.dartlang.org/angular/guide/security

PR #270, opened 2016-12-20
[ng] security guide

chalin:
This adds a security page + sample code but **does not** add the page to the sidenav.

This is a port of angular.io PR https://github.com/angular/angular.io/pull/1813, so that we don't lose the changes. Updates have been made to the prose, incorporating the latest TS-side changes (so that the chapter followed doc guidelines).

Note that a few pending remarks associated https://github.com/angular/angular.io/pull/1813 still need to be assessed/addressed.

Contributes to #263

- [ ] Create chapter prose.
- [ ] Create chapter example source.
- [x] Deploy example under `angular-examples`.


kwalrath:
@matanlurey would you be able to review this?

chalin:
@madsager @kwalrath - the page is now hosted here: https://angulardart-org-dev.firebaseapp.com/angular/guide/security

kwalrath:
I'll wait for Mads/Matan to take a look before I do.

public/docs/_examples/security/e2e-spec.ts
```diff
@@ -0,0 +1,37 @@
+'use strict'; // necessary for es6 output in node
```
matanlurey: Oops, is this file intentional?

chalin: Yes. At the moment, to ensure we have continued e2e test coverage, we are using the ng.io e2e protractor based tests.

src/angular/guide/security.jade
```diff
@@ -11,4 +11,24 @@ nextpage:
   url: /angular/guide/structural-directives
 ---
 //- FilePath: src/angular/guide/security.jade
-include /_jade/_ts-temp
```
kwalrath: @matanlurey where should emails about vulnerabilities go?

(The doc currently says "Email us at security@angular.io")

src/angular/guide/security.jade
```diff
@@ -11,4 +11,24 @@ nextpage:
...
\ No newline at end of file
+extends /_jade/ts/_cache/guide/security.jade
```
kwalrath: The live example & view source links don't work, of course. Could you publish those before we merge this PR?

kwalrath: The "Security Risk" link doesn't work—there is no #bypass-security-apis in this doc. Just remove the link?

kwalrath: Once this doc is ready, we should provide an online or PDF version of this page for Matan to review. I suspect there are some Dart differences. For example:

* The text in "Sanitization and security contexts" and "Sanitization example" might need to change because dart:html handles some of the sanitation.
* The wording/explanation for "Avoid direct use of the DOM APIs" should probably be tweaked to acknowledge dart:html.
* "Offline template compiler" sounds un-Darty, and the explanation links to the (non-existent for Dart) Dynamic Forms Cookbook.

public/docs/_examples/security/dart/lib/inner_html_binding_component.dart
```diff
@@ -0,0 +1,11 @@
+// #docregion
+import 'package:angular2/core.dart';
+
+@Component(
+    selector: 'inner-html-binding',
+    templateUrl: 'inner_html_binding_component.html')
+// #docregion class
+class InnerHtmlBindingComponent {
+  // E.g. a user/attacker controlled value from a URL.
```
kwalrath: E.g. -> For example,
attacker controlled -> attacker-controlled

(to match improvements in TS version)

src/angular/guide/security.jade
kwalrath: The last line of the page links to the non-existent bypass-security-apis section. It also mentions the bypassSecurityTrust methods for the first time. If we really do want to mention those methods (@matanlurey, do we?), we should probably link to the [DomSanitizationService API doc](https://webdev.dartlang.org/angular/api/angular2.security/DomSanitizationService-class) (which, btw, has an unclosed link early in its description).

src/angular/guide/security.jade
chalin: As an example, the [Google Application Security page](https://www.google.com/about/appsecurity/) suggests that Chrome users simply "_use the Chromium bug tracker_". I suggest that we do the same and refer users to the **AngularDart issue tracker**. WDYT?

src/angular/guide/security.jade
chalin: The example is now live.

src/angular/guide/security.jade
kwalrath: That sounds reasonable. Initially I hesitated because... security. (E.g., the Chromium bug tracker doesn't initially publish bugs that are marked as security bugs.) But I reckon people with the know-how to report a security bug will be careful about divulging too much information.

chalin: Right. The fact that the Chromium bug tracker can have private issues (which GH does not) does add extra protection in case too much _is_ divulged in the issue report.

From the Google App Security page:
> If you believe you have discovered a vulnerability in a Google product or have a security incident to report, go to http://goo.gl/vulnz 

How about using this link until an alternative is proposed? (I guess that it isn't much better than using security@angular.io). If not, I agree that it is probably best to leave the recommendation as "Email us at security@angular.io".

src/angular/guide/security.jade
chalin: > The "Security Risk" link doesn't work
I've removed the link

chalin: > The "Security Risk" link doesn't work

The fragment being linked to now exists.

public/docs/_examples/security/dart/lib/inner_html_binding_component.dart
chalin: Fixed

chalin:
Added new section on "Trusting safe values".

Staged at https://angulardart-org-dev.firebaseapp.com/angular/guide/security

New section is [here](https://angulardart-org-dev.firebaseapp.com/angular/guide/security#trusting-safe-values).

src/angular/guide/security.jade
kwalrath: Let's keep it as-is, for now. Any Dart security issues are likely to be in the TS version, too.

src/angular/guide/security.jade
kwalrath: We should mention the words "Security risk" in the "Trusting safe values" section, so it's clearer that they've gotten to the right section, but since the TS version doesn't, it's not a blocking issue.

src/angular/_jade/ts/_cache/guide/security.jade
```diff
@@ -38,7 +38,7 @@ h2#best-practices Best practices
   * **Keep current with the latest Angular library releases.**
   We regularly update our Angular libraries, and these updates may fix security defects discovered in
   previous versions. Check the Angular [change
-  log](https://github.com/angular/angular/blob/master/CHANGELOG.md) for security-related updates.
+  log](https://github.com/dart-lang/angular2/blob/master/CHANGELOG.md) for security-related updates.
```
chalin: Right. This file is our (cached) copy of the TS security page. The changes here will not affect ng.io TS files. It just means that we'll have this small diff when we compare to the corresponding TS file.

chalin:
Updates made:
- Removed sentence referring to the cookbook.
- Fixed the sentence referring to "Security Risk" and the section on "Trusting safe values".

As for mention of `dart:html`, that will likely be addressed once the last (currently commented out) section is incorporated into the guide.
> ... he's too busy, we might be better off publishing this draft than waiting.

I'd be in favor of that.

kwalrath:
Should be auto-published in a few minutes.

PR #280, opened 2016-12-28
Angular homepage

kwalrath:
Staged version: https://kw-webdev-dartlang-1.firebaseapp.com/angular

You might need to clear the cache and/or open it in an incognito window to avoid redirecting to /angular/guide.


src/angular/index.md
```diff
@@ -0,0 +1,51 @@
+---
+layout: angular
+title: About AngularDart
+description: Angular + Dart = productivity, performance, and stability
+toc: true
```
chalin: I'd suggest `toc: false`; the page is so short (even in a narrow view), I'm not sure that adding a(n autogenerated) TOC has much value.

src/_data/side-nav.yml
```diff
@@ -3,10 +3,10 @@
   - title: Get Started
     permalink: /guides/get-started
 
-- title: Angular Docs
+- title: Angular
   children:
-    - title: Quickstart
-      permalink: /angular/quickstart
+    - title: About AngularDart
```
chalin: How about just "About" in the sidenav?

src/angular/index.md
kwalrath: Yep, that was an oversight.

src/_data/side-nav.yml
kwalrath: I like it!

kwalrath:
I'm inclined to ship this. We can iterate, of course.

PR #307, opened 2017-01-11
Add more Angular

kwalrath:
Here's a try at adding more obvious Angular links to the webdev homepage (with some ripples extending to the Angular homepage). Staged at:

https://kw-webdev-dartlang-1.firebaseapp.com


src/index.html
```diff
@@ -9,6 +9,8 @@
     <div class="content">
       <h2>Build web apps in a modern language, using<br>stable libraries and solid tools.</h2>
       <a href="/guides/get-started" class="btn btn-default btn-lg">Get Started</a>
+      &nbsp;&nbsp;
```
kevmoo: If we want to get fancy, we could update all of these URLs to

`/angular/?utm_source=home&utm_campaign=more_ng&utm_medium=link`

Then we can see in analytics how much additional traction we get from these changes

kevmoo: although this would create address-bar noise.

Thoughts?

kwalrath: address-bar noise? oh, you mean readers would see that URL?

Yeah, I'm inclined not to do it for that reason. I can put a note in the analytics about adding this change, so we can easily know which dates to use to compare before-after numbers.

kevmoo: Cool

Issue #279, opened 2016-12-28
weird flow in multi-line bullet link


kwalrath:
Looking at my local version of the Angular homepage-to-be (http://localhost:4001/angular), if the screen is at a size where the bulleted list of videos has line breaks in individual items, the formatting is terrible. The bullet appears on the last line instead of the first; the offsite link icon is on the first line instead of the second, as well as often being too far to the right. Here's an example:

![image](https://cloud.githubusercontent.com/assets/2164483/21533265/a0504c1e-cd0d-11e6-8019-43b06eeff570.png)

It looks correct when the bulleted list items are single lines:

![image](https://cloud.githubusercontent.com/assets/2164483/21533295/00631be0-cd0e-11e6-9be1-859c2e3c420c.png)

kwalrath:
This happens in www too, and in both Chrome & Firefox. It has nothing to do with the column styles that happen to surround the bulleted lists in the Angular homepage. 

As the following screenshot shows, the bullet is correctly placed if the external link icon isn't present.

![image](https://cloud.githubusercontent.com/assets/2164483/21534063/75ed33f2-cd16-11e6-9e8a-9593c73fd6be.png)

chalin:
A first question to ask is whether we really want to keep the _external link_ icon? I'm not sure that it adds much value. That simplest solution might be to just drop the insertion of the icon.

kwalrath:
We added the external link icon in response to a request (long ago), so I hesitate to drop it. If we were to drop it, then we might want to add context to many links, so it's clear how local/far they are—same page, same site, other dartlang.org site, other site entirely. (This is something we might want do anyway, but it's not quick/easy.)

PR #342, opened 2017-01-27
Remove 'with Dart' from button on /angular

filiph:
IMHO, a button that says 'Get started *with Dart*' looks scary and irrelevant if you're coming for AngularDart. The linked page ends up with building an AngularDart app anyway, so it's really 'Get started with AngularDart', so we can lose the 'with Dart'.


kwalrath:
Looking at the new version, I wonder if part of the reason I added "with Dart" was just to make the buttons look at little more balanced. (And also not to imply that you needed Dart knowledge to get started.)

kwalrath:
Also... I need to work on #279.

PR #344, opened 2017-01-28
Add Angular version page

kwalrath:
A version string (and link to new version page) is now in the title of every page.
Also changed the Angular shield in the title to link to /angular.
And made misc other nav/word tweaks.

Staged:
https://kw-webdev-dartlang-1.firebaseapp.com/angular/version


src/_data/side-nav.yml
```diff
@@ -81,15 +81,12 @@
...
-  children:
-  - title: "Overview"
-    permalink: /guides/web-programming
-
 - title: "Resources"
   children:
   - title: "Dart Tools for the Web"
     permalink: /tools
+  - title: "Web Libraries"
+    permalink: /guides/web-programming
```
chalin: Good move to put this under Resources.

src/_includes/navigation-main.html
```diff
@@ -5,7 +5,7 @@
   </a>
   <ul>
     <li><a href="/guides/get-started">Get Started</a></li>
-    <li><a href="/guides/web-programming">Libraries</a></li>
+    <li><a href="/angular">Angular</a></li>
```
chalin: Also a good move to emphasize Angular in the main nav.

src/_layouts/angular.html
```diff
@@ -11,7 +11,12 @@
           {% include banner.html %}
           {% include navigation-sub.html %}
           {% include breadcrumbs.html %}
-          <h1>{% if page.long-title %}{{ page.long-title }}{% else %}{{ page.title }}{% endif %}</h1>
+          <div id="title">
+            <h1>
+              <a href="/angular"><img src="/angular/images/angulardart.svg"></a>
+              {% if page.long-title %}{{ page.long-title }}{% else %}{{ page.title }}{% endif %} </h1>
```
chalin: You've removed all `long-title` entries, so this can be simplified to just `{{ page.title }}`.

src/_layouts/angular.html
```diff
@@ -11,7 +11,12 @@
           {% include banner.html %}
           {% include navigation-sub.html %}
           {% include breadcrumbs.html %}
-          <h1>{% if page.long-title %}{{ page.long-title }}{% else %}{{ page.title }}{% endif %}</h1>
+          <div id="title">
```
chalin: It would be preferable, IMHO, to make use of a CSS class instead of an element id here --- maybe `ng-page-title`.

src/_layouts/angular.html
```diff
@@ -11,7 +11,12 @@
...
+            <h1>
+              <a href="/angular"><img src="/angular/images/angulardart.svg"></a>
+              {% if page.long-title %}{{ page.long-title }}{% else %}{{ page.title }}{% endif %} </h1>
+            <p id="version-in-title"> <a href="/angular/version">2.2</a> </p>
```
chalin: It would probably be best to define and make use of global (site-wide) definitions named, say: `ng-stable-version`, `ng-dev-version`. Then automating the update of these version numbers would be slightly easier.

src/angular/index.md
```diff
@@ -12,17 +12,14 @@ mission-critical apps that bring in much of Google’s revenue.
...
 
-  <em>Latest version:
-  <b><ngio-cheatsheet
-      src='/angular/cheatsheet.json' version-only></ngio-cheatsheet></b>
-  (<a href="https://pub.dartlang.org/packages/%61ngular2/versions/3.0.0-alpha#changelog"
-      class="no-automatic-external"
-      target="_blank" rel="noopener">changelog</a>)</em>
+  <em>
+    <a href="/angular/version">Latest stable version:</a>
+    <b><ngio-cheatsheet src='/angular/cheatsheet.json' version-only></ngio-cheatsheet></b>
```
chalin: If we opt for use of a global `ng-stable-version`, then we can use that here instead of the `<ngio-cheatsheet version-only>` directive.

src/angular/version.md
```diff
@@ -0,0 +1,71 @@
...
+Each example in the AngularDart documentation has a repo under the GitHub organization
+[angular-examples](https://github.com/angular-examples).
+These example repos are generated from the dart-lang/site-webdev repo,
+using files under the
+[public/docs/_examples](https://github.com/dart-lang/site-webdev/tree/master/public/docs/_examples)
+directory.
+
+We intend to create 3.0 versions of all the examples,
+placing them in a branch of the dart-lang/site-webdev repo.
+Once 3.0 is stable, we'll update the example repos.
```
chalin: Nice version page.

chalin:
As for the page title CSS, I'd recommend sticking to the use of a background image. The old CSS resulted in "clean" title wrapping for narrow displays, like this:
> screen shot 2017-01-28 at 10 58 00 am

On the other hand, the new titles look like this:
> screen shot 2017-01-28 at 10 57 44 am

We can get the version number to appear _and_ get proper title wrapping with the proper CSS.

src/_layouts/angular.html
kwalrath: I'll try that.

src/_layouts/angular.html
kwalrath: We had long-titles in the old site, so I think we might need them here too, eventually.

src/_layouts/angular.html
kwalrath: Sounds good.

src/angular/version.md
kwalrath: Thanks!

src/_layouts/angular.html
chalin: The "long-title" was a feature I added so that we could avoid displaying a muddled version number in the sidebar when showing, say, the Cheatsheet tile.

But, if you'd like to keep the feature that is fine too.

kwalrath: I remember adding a similar feature to the old www.dartlang.org because sometimes you wanted a longer title on the page, but a shorter title in the social snippet. Or maybe it was the other way around. I wonder if we have it implemented twice.

chalin: IIRC, there wasn't any other field that influenced the titles, but there were different fields that seemed to impact the page summary/banner.

kwalrath:
> As for the page title CSS, I'd recommend sticking to the use of a background image...
> We can get the version number to appear _and_ get proper title wrapping with the proper CSS.

I wanted clicking the Angular shield to take you to the Angular homepage. Is there any way to do that with a background image?

kwalrath:
I've made most of the changes Patrice suggested, none of which should be visible (but they'll make updating the site easier).

The one thing I didn't do was to fix the heading wrap. I experimented with the following, but it didn't work and I didn't want to spend _too_ much time on it.

```
.ng-page-title {
  margin-left: 3.1em;
  text-indent: -5.3em;
}
```

PR #355, opened 2017-02-10
[ng] ng.io sync mainly of images and script files

chalin:
Contributes to #357.


.gitignore
```diff
@@ -44,5 +44,6 @@ pubspec.lock
 a2docs.css
 # !/public/docs/_examples/
 # /public/docs/_examples/node_modules
-# /public/docs/_examples/**/a2docs.css
+/public/docs/_examples/*/dart/web/a2docs.css
+/public/docs/_examples/*/dart/web/styles.css
```
kwalrath: Do we really want to ignore styles.css? Oh! Is this so you can have a local copy of the styles.css file without the risk of having different styles.css files?

public/docs/_examples/_boilerplate/plnkr.json
```diff
@@ -0,0 +1,10 @@
+{
```
kwalrath: Do we really need this file?

.gitignore
chalin: Right.

public/docs/_examples/_boilerplate/plnkr.json
chalin: Not really, but I've been just sync'ing the entire _boilerplate folder for now. (The file just gets ignored in our build. I can eventually refine the sync filter ... on the next round?)

chalin:
Most of the next images are for guide changes that are forthcoming (or hopefully new guides that we should also be adding soon :).

PR #356, opened 2017-02-10
Clarify the Angular doc path

### LINKED ISSUE #217: Improve the Angular quickstart
Improve the [quickstart](https://webdev.dartlang.org/angular/quickstart) in two parts:

1. Incorporate the changes from https://github.com/angular/angular.io/pull/2946.
2. Once we've separated, make further edits (copy edits, further Dartification, ...). Consider the revised [Get Started](https://webdev.dartlang.org/guides/get-started) guide; maybe we don't actually need two pages?

kwalrath:
Fixes #217 

Twice my repo broke (fatal: cache-tree: internal error) while I prepped this, but third time's the charm.

Staged at: https://kw-webdev-dartlang-1.firebaseapp.com/angular

The most noticeable changes are in https://kw-webdev-dartlang-1.firebaseapp.com/angular/tutorial/toh-pt1


src/angular/guide/index.jade
```diff
@@ -11,8 +11,81 @@ nextpage:
...
+
+  ## Reference pages
+
+  * The [Glossary](glossary) defines terms that Angular developers should know.
+  * The [Cheat Sheet](cheatsheet) lists Angular syntax for common scenarios.
+  * The [API Reference](../api) is the authority on every public-facing member of the Angular libraries.
+
+  ## Feedback
+
+  We welcome feedback!
```
chalin: Consider changing the section title from "Feedback" to "We welcome feedback!" and then just drop this paragraph.

src/angular/guide/setup.jade
```diff
@@ -57,8 +57,11 @@ a#webstorm
   This only happens once. After that, you'll be able to use WebStorm for the usual IDE tasks,
   including running the app.
 
+  If WebStorm shows errors for the app, it probably didn't get the dependencies for your app.
+  The workaround is to right-click `pubspec.yaml` and choose **Pub: Get Dependencies.**
```
chalin: Consider: "The workaround is to ..." --> "To fix this ..."
("workaround" makes it sound like this is a persistent issue with WS which JB is choosing not to fix).

src/angular/tutorial/toh-pt1.jade
```diff
@@ -13,42 +13,25 @@ nextpage:
 include ../_util-fns
 
 :marked
-  Every story starts somewhere. Our story starts where the [QuickStart](../quickstart.html) ends.
+  This tutorial starts with about the simplest Angular app possible.
```
chalin: Consider dropping "about". It makes the sentence difficult to read. If you feel the claim becomes too strong then maybe: "This tutorial starts with a very simple Angular app."

src/angular/tutorial/toh-pt1.jade
```diff
@@ -57,6 +40,32 @@ code-example(language="sh" class="code-shell").
   This command runs the compiler in watch mode, starts the server,
   and keeps the app running while we continue to build the Tour of Heroes.
 
+
+.l-sub-section
+  :marked
+    ### About Angular applications
```
chalin: This side-note doesn't quite work for me. I'll make a suggestion as a top level comment to this PR since it crosses page boundaries.

chalin:
In the current edit:
- The Angular landing page has a "Quick start" button which leads to the "Get Started" page. But this page actually **_doesn't_** make use of the "Angular QuickStart Example" template.
- The tutorial (part 1) says to use the "Angular QuickStart Example" template, which most readers might assume would have been discussed in the "Get Started" page (target to the "Quick start button" :).

I think that we can improve on this w/o making too many more structural changes. I'd suggest that:

- We rename the stagehand template "Angular QuickStart Example" to "Simple Angular Application" and, to be consistent with other "Simple *" templates, edit the template description to read "a bare-bones Angular web app".
- Rather than merge the QuickStart prose into a side-note of the toh-1 page, I would create a separate page for it as part of the tutorial, maybe "App Starter":

  >&nbsp;Introduction
  >&nbsp;0. App Starter
  >&nbsp;1. The Hero Editor
  >&nbsp;2. Master/Detail
  >&nbsp;...

  The "App Starter" page can now contain the equivalent of the QuickStart content in addition to the toh-1 "Keep the app running". This way, "The Hero Editor" section is actually focused on introducing the first hero-specific elements of the app.

Having a distinct page like "App Starter" also means that we have a target for links that used to point to the QuickStart (which I assume you will be deleting). Pages like attribute-directives, displaying-data, template-syntax currently have links to the QuickStart page.

src/angular/guide/index.jade
kwalrath: Our style (which I finally found corroborated in https://docs.google.com/document/d/1ETcX_38vkR8NkGyKqhWoCnPDg3WuLw8UifO5TGCHbSo/edit#heading=h.76serc5rcimf) is to precede a list with an introductory sentence.

This was not the best sentence, though. I'll try to write something a little more helpful.

chalin: I think that this might qualify for an exception :).

See how the preceding section has no such introductory paragraph (and the section in only a list).

kwalrath: I didn't fix every editorial mishap in this page. :)

chalin: Yes, I know the feeling :)

kwalrath:
I agree, renaming the "QuickStart" template seems like the way to go. What do you think of making the Setup page the one that everything refers to?

chalin:
I prefer the separation of concerns that we'd have with a "Setup" separate from an "App Starter". The "Setup" has details about setting up development environment. The "App Starter", as the base step of the tutorial, informs us of what the basic parts of an app are (which doesn't really fit under a section entitled "Setup" IMHO).

kwalrath:
Yep, I think you're right. I'm not sure the first page of the tutorial is where I want to refer everyone for basic setup info, but it does seem like the right place to discuss the basic sample.

kwalrath:
I've added the new page (https://kw-webdev-dartlang-1.firebaseapp.com/angular/tutorial/toh-pt0) but haven't expunged mentions of "QuickStart" everywhere yet. If this flow works for you, @chalin, I should be able to finish up on Monday.

chalin:
Yes! Very nice "Starter App" page.

kwalrath:
I've updated everything so it's in a coherent state, except for a couple of things I'd like to do separately:
* rename the QuickStart template
* create a https://github.com/angular-examples/toh-5 repo

What do you think, @chalin, is this good to go (and iterate on)?

Staged at URLs like these: 
https://kw-webdev-dartlang-1.firebaseapp.com/angular/tutorial/toh-pt0
https://kw-webdev-dartlang-1.firebaseapp.com/angular/guide/displaying-data

src/angular/guide/displaying-data.jade
```diff
@@ -11,24 +11,266 @@ nextpage:
...
+  :marked
+    The <live-example></live-example> demonstrates all of the syntax and code
+    snippets described in this chapter.
+
+.l-main-section#interpolation
+:marked
+  ## Showing component properties with interpolation
+  The easiest way to display a component property
+  is to bind the property name through interpolation.
+  With interpolation, we put the property name in the view template, enclosed in double curly braces: `{{myHero}}`.
```
chalin: Using Jade `extends` isolates us from having the Liquid processor gobble up interpolation synatx `{{ … }}` that might appear in the extended file. Now that the Jade is all in _this_ file, you'll need to escape interpolation syntax, like the `{{myHero}}`.

src/angular/guide/displaying-data.jade
```diff
@@ -11,24 +11,266 @@ nextpage:
...
+
+  Let's build a small illustrative example together.
+
+  Create a starter app in a directory named
+  <ngio-ex path="displaying-data"></ngio-ex>,
```
chalin: When directives appear at the start of the line like this, because Jade doesn't recognize this as an "inline" HTML element, it assumes that it is a block element, and hence puts it in a separate paragraph. You'll notice that the above is rendered as:
> Create a starter app in a directory named
>
> displaying_data, following the steps in Setup for Development.

Actually, you can just replace all of line 47 with the folder name:
```
  `displaying_data`
```

src/angular/guide/index.jade
```diff
@@ -11,8 +11,82 @@ nextpage:
...
 ---
 //- FilePath: src/angular/guide/index.jade
-extends /_jade/ts/_cache/guide/index.jade
+include ../_util-fns
+- var guideData = {}; // public.docsdart.latest.guide._data;
+- var advancedLandingPage = 'attribute-directives'; // '';
+- for(var page in guideData) {
+-   if (!guideData[page].basics && !guideData[page].hide) { advancedLandingPage = page; break; }
+- }
+- var advancedUrl = './' + advancedLandingPage
```
chalin: You can replace lines 15-20 with just the following, (since we don't be diffing with the TS _cache file version anymore):
> ```- var advancedUrl = './attribute-directives'```

A clean solution might be to add the name of the first advanced page to the site `_config.yml` file, and then use that definition directly in the URL below (and drop the `advancedUrl` variable altogether). But making the change above will do for now.

chalin: You can replace lines 15-20 with just the following, (since we don't be diffing with the TS _cache file version anymore):
> ```- var advancedUrl = './attribute-directives'```

A clean solution might be to add the name of the first advanced page to the site `_config.yml` file, and then use that definition directly in the URL below (and drop the `advancedUrl` variable altogether). But making the change above will do for now.

src/angular/tutorial/toh-pt0.jade
```diff
@@ -0,0 +1,147 @@
+---
+layout: angular
+title: The Starter App
+description: We look at a simple Angular app
```
chalin: Should we avoid "We" in the description?

src/angular/tutorial/toh-pt0.jade
```diff
@@ -0,0 +1,147 @@
...
+  url: /angular/tutorial
+nextpage:
+  title: The Hero Editor
+  url: /angular/tutorial/toh-pt1
+---
+//- FilePath: src/angular/tutorial/toh-pt1.jade
+include ../_util-fns
+
+:marked
+  This tutorial starts with a very simple, boilerplate Angular app.
```
chalin: "boilerplate" --> "simple"
or "bare bones"?

src/angular/guide/setup.jade
```diff
@@ -76,99 +86,37 @@ a#cli
   `pub global activate stagehand`
   1. Create a directory for your project:
   <br>
-  `mkdir angular_quickstart; cd angular_quickstart`
+  `mkdir my_app; cd my_app`
```
chalin: Somehow the `my_app` name doesn't feel quite right because it matches the selector name of the `AppComponent`.

How about `angular_project` or `angular_app`?

chalin:
Also, regarding the toh-0 repo, I think that it would be preferable to give such a repo a more meaningful name, like `starter-app` or just `starter`.  I'd even favor keeping `quickstart` over creating a new repo named `toh-0`, though giving it a new name like `starter` has appeal.

src/angular/guide/displaying-data.jade
kwalrath: Ah shoot, I caught this earlier but not here. Good catch!

chalin:
Btw, would you mind keeping the cache files?  It will make it easier to find out what has changed on the TS side in between our doc sync points?

kwalrath:
I'm OK with keeping the cache files if you want, but I won't use them at all. Should we restore the files I deleted earlier? If so, can you deal with that?

chalin:
> I won't use them at all

That's fine.
> can you deal with that?

Yes

src/angular/tutorial/toh-pt0.jade
kwalrath: They all start with "We". I'll fix them all later.

src/angular/guide/setup.jade
kwalrath: I'll just use `angular_quickstart`, to match what's in the WebStorm instructions (for now).

kwalrath:
OK, I think it's ready to go. Agreed, @chalin?

src/angular/guide/displaying-data.jade
```diff
@@ -39,12 +39,11 @@ figure.image-display
...
   The easiest way to display a component property
   is to bind the property name through interpolation.
-  With interpolation, we put the property name in the view template, enclosed in double curly braces: `{{myHero}}`.
+  With interpolation, we put the property name in the view template, enclosed in double curly braces: {% raw %}`{{myHero}}`{% endraw %}.
 
   Let's build a small illustrative example together.
 
-  Create a starter app in a directory named
-  <ngio-ex path="displaying-data"></ngio-ex>,
+  Create a starter app in a directory named `displaying-data`,
```
chalin: The folder name should be `displaying_data` (with an underscore rather than a dash).

kwalrath: Gah! I can't believe I messed that up. Good thing you're checking.

PR #360, opened 2017-02-10
Add Twitter follow button for @AngularDart

filiph:
This adds the official follow button on the /angular landing page, next to news.

Twitters implementation flashes the naked text (follow @AngularDart) before the script can load. Hopefully, most users have the twitter-hosted script loaded, so there’s no flash.


kwalrath:
I see that flash every time I (re)load the page... it's pretty annoying. Can we do something to make it never show up? Or maybe style it ourselves so the change isn't so obvious?

kwalrath:
Also... we need to post there more.

chalin:
If you change the text to `"Follow @AngularDart"` and set the font size to "small", the transition is a lot smoother (since the font size almost matches): i.e., from
> screen shot 2017-02-11 at 12 56 33 pm

to
> screen shot 2017-02-11 at 12 56 47 pm

filiph:
Great points, @kwalrath and @chalin. I inlined the twitter styles. It's now just a regular link to the Twitter page. But also loads much faster, with no FUC.

PTAL.

PS: Having more content on @AngularDart is something I sure need to solve.

chalin:
@filiph, have you staged this somewhere?

I've done a clean and full rebuild of the site and the New heading looks like this:
> screen shot 2017-02-14 at 9 10 35 am

Was the intent to loose the button?

filiph:
No, that's not right. I didn't stage it and won't get to it for some time.

It looks like the new CSS isn't applying at all. Can you make sure it's
built and refreshed in the browser? I had a similar issue while building it.

2017-02-14 9:13 GMT-08:00 Patrice Chalin <notifications@github.com>:

chalin:
I was careful and relaunched an new incognito window, but I'll try again later today.

kwalrath:
I saw _nothing_ the first time I loaded the page. (Maybe the text was white?) When I reloaded, I saw the styled button. LGTM.

chalin:
Great, at least one of us is seeing the right result :)

kwalrath:
I tried again. At first I saw nothing, and (after reloading) then saw the big-font **Follow @AngularDart** again. :/ Reloading again, I finally saw the button.

After that, I always see the button, whether in incognito or in another browser. So I guess it's fixed? Let's see what happens. :)

kwalrath:
The change is visible at https://angulardart-org-dev.firebaseapp.com/angular.

It's not visible at webdev.dartlang.org yet because we're working on 3.0 docs, and not deploying webdev until 3.0 is out.

PR #413, opened 2017-02-27
Update get-started.md

anders-sandholm:
fix based on feedback from users


kwalrath:
Thanks, @anders-sandholm!

kwalrath:
The fix is visible at https://angulardart-org-dev.firebaseapp.com/guides/get-started.

It's not visible at webdev.dartlang.org yet because we're working on 3.0 docs, and not deploying webdev until 3.0 is out. (If it's urgent to deploy the fix, let me know.)

PR #432, opened 2017-03-06
[ng] ng.io sync: structural directives

chalin:


kwalrath:
It looks like you changed each occurrence of `<ng-container>` to `<template>`. I'm not sure that works here:

> The `<template>` is a syntax element recognized by the Angular parser. It's not a directive, component, class, or interface. It's more like the curly braces in a Dart `if`-block...

Should we just delete that paragraph and the following example?

There might be other places where substituting ng-container for template requires additional text changes.

kwalrath:
Once you fix any problems with `<template>`, I'm ok with publishing this, with the understanding that it'll need a copy edit later.

chalin:
AFAIK, that sentence is also accurate for `<template>`, so I'd keep it.

> There might be other places where substituting ng-container for template requires additional text changes.

I was careful in making the change to ensure that the surrounding text remained valid.

So I think that this is ready to go then?

chalin:
Here is a link to the `<template>` element: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/template

PR #442, opened 2017-03-09
[ng] ng.io sync: template syntax, code and prose updates

chalin:
Note: a link check skip rule was added, pending the merge of the Structural Directives PR.

Staged at https://angulardart-org-dev2.firebaseapp.com/angular/guide/template-syntax


src/angular/guide/template-syntax.jade
```diff
@@ -22,101 +22,131 @@ include ../_util-fns
...
 
-  Let’s find out what it takes to write a template for our view. We’ll cover these basic elements of template syntax:
+  You may be familiar with the component/template duality from your experience with model-view-controller (MVC) or model-view-viewmodel (MVVM).
+  In Angular, the component plays the part of the controller/viewmodel, and the template represents the view.
 
-  * [HTML](#html)
-  * [Interpolation](#interpolation)
+a#toc
+:marked
+  ### Table of contents
```
chalin: IMHO, the two TOCs are complementary.
- The in-page nav has the full details.
- The side nav TOC is conveniently terse and always there; it also provides "spacial" context to let the reader know where they are in the page (though it doesn't _always_ work, but that's an issue for another day).

The current autogenerated TOC is overly brittle. Ideally we should replace with with an Angular directive, but it isn't something that I think we should do quite yet.

kwalrath: I really dislike duplicating the TOC. I'd rather have a textual description, with links, than 2 TOCs.

kwalrath: To be more specific than "really dislike", I think that the two TOCs might get out of sync, and that even if they're in sync, it looks like a bug to have the TOC visible twice.

It just feels wrong to have the same text visible in two places.

chalin: > Table of contents -> Contents

Done

> For pages like this with an in-page TOC we should turn off the TOC in the side nav.

Done

src/angular/guide/template-syntax.jade
```diff
@@ -22,101 +22,131 @@ include ../_util-fns
 - var __objectAsMap = '<b><a href="' + mapApiRef + '">Map</a></b>'
 
 :marked
-  Our Angular application manages what the user sees and can do, achieving this through the interaction of a Component class instance (the *component*) and its user-facing template.
+  {{page.description}}
```
chalin: That looks good. As agreed, I've done something a little simpler for this first PR. We can look into a general solution after.

PR #405, opened 2017-02-23
Update get-started.md

anders-sandholm:
I was missing this info when going through the process.

PR #446, opened 2017-03-13
Update get-started.md

kwalrath:
Alternate implementation of #405.

I decided not to get too specific, since the doc in general tries to be brief and cover the best case. (If we change that, we'll want to have more details here and elsewhere.)


anders-sandholm:
Thanks for fixing. Hopefully this will only be a stop-gap solution until we're able to pick up the paths automatically.

PR #447, opened 2017-03-13
Add info about transformer order

### LINKED ISSUE #254: emphasize that dart_to_js_script_rewriter transformer must be after angular2 transformer
_From @kasperpeulen on December 22, 2015 14:25_

@kwalrath 
In https://angular.io/docs/dart/latest/quickstart.html it could be empasized that dart_to_js_script_rewriter transformer must be after angular2 transformer. If this is not done, templates break:

image

_Copied from original issue: angular/angular.io#600_

kwalrath:
Fixes #254

/cc @kasperpeulen


kwalrath:
Yeah, I was tempted to say make it last, but I wasn't sure enough to say so.

chalin:
The prose you added is just fine.

PR #459, opened 2017-03-15
[ng] enable component-styles e2e

chalin:
**Built atop of #457**.

- Enable e2e for `component-styles`, by disabling single test known to fail (ng fix pending).
- Fix glob pattern so that we stop syncing `style-guide` and `styleguide` e2e files.


kwalrath:
What's the bug # for the test we're skipping?

chalin:
> What's the bug # for the test we're skipping?

Yes.

chalin:
Since you approved, and I answered your question, I'll go ahead and merge.

Issue #261, opened 2016-12-16
Add e2e tests for Developer Guide Dart examples


chalin:
_From @chalin on May 25, 2016 18:23_

- [x] attribute-directives - angular/angular.io#1765
- [x] component-styles - #459
- [x] hierarchical-dependency-injection
- [x] server-communication - angular/angular.io#1423
- [x] lifecycle-hooks - angular/angular.io#1768
- [x] pipes - angular/angular.io#1353
- [x] router - #460
- [x] structural-directives

There are no `testing` and `configuration` chapters yet. Contributes to angular/angular.io#1521.

_Copied from original issue: angular/angular.io#1523_

PR #460, opened 2017-03-16
[ng] Enable router E2E tests

chalin:
- Await for alert dialog and use it in test for canceling crisis detail change.
- Tests now use async/await. It makes them easier to read.
- Adapted tests to Dart sample configuration.

Contributes to #261


kwalrath:
I won't try to understand these, but since they work...

PR #473, opened 2017-03-21
[ng] chore(travis): run current suite of e2e tests

chalin:
- run `examples-install.sh` for e2e
- run e2e as `--fast`
- start `xvfb`
- use trusty env so that we get the latest version of chrome (older versions are incompatible with protractor 4)


kwalrath:
I'm not pretending to understand the testing code, but I do understand that (1) this works and (2) this doesn't slow down deployments. So we're good to go.

PR #486, opened 2017-03-25
[ng] ngio toh-5 sync and other page updates

chalin:


gulp/ngio-get.js
```diff
@@ -136,7 +141,7 @@ module.exports = function (gulp, plugins, config) {
   gulp.task('_get-extra-dart', () => {
     const baseDir = path.join(angulario, 'public/docs/dart/latest');
     return gulp.src([
-      `${baseDir}/_util-fns.jade`,
+      `!${baseDir}/_util-fns.jade`, // 2017-03-24 stop syncing _util-fns.jade
```
kwalrath: Just curious, why is this `!` instead of `//`?

kwalrath: (since the next line uses `//`)

chalin: Commenting it out would actually have been ok too (and probably better).

chalin: Fixed

PR #487, opened 2017-03-27
fix link scroll adjustment for ng and non-ng pages

### LINKED ISSUE #315: [ng] docs doesn't scroll to the appropriate section
1. Open new tab in Chrome
2. Paste any url which has the hash part (this type of url is generated when clicking through the right navigation - for example https://webdev.dartlang.org/angular/guide/template-syntax#!#two-way-binding-with-ngmodel (or any other))
3. It doesn't scroll to the appropriate part of the page

Seems to be problematic only in Angular docs.

### LINKED ISSUE #483: in-page links scroll up to top
I just noticed today that when I link to https://webdev.dartlang.org/guides/get-started#2-install-dart, the page starts out showing the section I want, but then scrolls up to the top.

This is in Chrome 59 canary, Chrome 56, and Safari 10.0.3.

It behaves fine in Firefox—great, even. It starts out below the header (and usual) and scrolls up just until you can see the section header. I suspect that's what used to happen in Chrome/Safari.

Same thing happens in other webdev pages:
https://webdev.dartlang.org/codelabs/angular2_components/2-easystart#--set-up-the-root-components-dart-file

However, www is fine (which makes me wonder if this is related to Angular CSS changes):
https://www.dartlang.org/guides/language/language-tour#operators

Angular pages are even more broken, not even starting at the section header, as reported in #315.

cc @filiph @chalin

chalin:
Fixes #315
Fixes #483

With this fix, link anchor offsets get handled in CSS using the same solution as that implemented for site-www. We also disable the scroll adjustments that used to be done via the `scroll-y-offset-element` directive (for angular doc pages). This way we have a uniform scheme that works for all pages angular and non-angular pages.

The offsetting is achieved via an "anchor" class which is set for header and anchor elements with id's. It is the `jekyll-toc` gem that adds this class, but only for pages for which the gem's functionality is enabled -- setting the Jekyll page property `toc: false` disables it.

Comments on individual changes:
- src/_assets/javascripts/main.js: remove on load function which attempted to adjust the y scroll position to compensate for the main nav. This failed to work on site-www and was commented out there too. Here, in addition, it also clashed with the ng directive.
- src/_assets/stylesheets/main.scss: define `anchor` class; cf. https://github.com/dart-lang/site-www/pull/154.
  - This also adds an octicon-link next to anchors or headers with ids
  - The supporting icon file is: src/_assets/icons/ic_link_black_24px.svg
- src/resources/js/directives/scroll-y-offset-element.js: don't adjust the y scroll offset for ng pages anymore. (Note that removing the directive entirely prevents ng hash paths (`#!#`) from being processed.)


kwalrath:
Looks great! I'll play around with the site a bit, but so far LGTM.

PR #510, opened 2017-04-01
chore(travis): only try to deploy for build tasks, not testing tasks

chalin:


chalin:
@kwalrath - our PR builds have been green, but commits to master have been failing because all tasks tried to deploy (as opposed to only the task that actually builds the site being activated for deployment. This fixes the issue.

kwalrath:
LGTM. Thanks for fixing this.

PR #516, opened 2017-04-04
[ng] toh-1: angular_test based tests

chalin:


kwalrath:
LGTM if the build completes successfully.

chalin:
Travis is happy now (it took it a few tries; at first there were pub server failures then Travis linux update failures).

PR #517, opened 2017-04-04
Implement Luke's UX suggestions

kwalrath:
(plus a few more tweaks)

Staged: https://kw-webdev-dartlang-1.firebaseapp.com


src/_data/side-nav.yml
```diff
@@ -3,7 +3,7 @@
   - title: Get Started
     permalink: /guides/get-started
 
-- title: Angular
+- title: Angular&nbsp;Dart
```
chalin: Why the `&nbsp` here (and on line 88) and not elsewhere in the edits done for this PR?

kwalrath: Because we need a space there (because of the ALL CAPS) but still wanted the words to always be together.

PR #548, opened 2017-04-13
[WIP][ng] Test page section cover toh-1 tests

chalin:
WIP. This version includes all toh-* tests in a single page.

Staged at: https://angulardart-org-dev.firebaseapp.com/angular/guide/testing#test-the-hero-editor-using-page-objects


kwalrath:
I'm not sure how useful it is to have the Testing page follow the TOH tutorial. Readers might be better served by a more task-oriented approach, rather than an example-oriented approach. We could still provide the full-testing source code for TOH, but as a bonus, not a focus.

PR #555, opened 2017-04-19
[ng] toh-3 component tests

chalin:


chalin:
Oops, forgot to enable the toh-3 tests on Travis. Just added that in a separate commit. Will wait for Travis to report green and merge if that is ok with you @kevmoo.

kevmoo:
👍

PR #557, opened 2017-04-19
[ng] toh-5, initial component tests

chalin:


chalin:
@kevmoo - this has useful tests for two of the app components. I'd say its ready for review.

PR #558, opened 2017-04-20
Follow-up on rename of GitHub repos for angular(2)_component(_example)

kevmoo:
This is only for GitHub-specific URLs.
The rename of the package angular2_component will happen later

See https://github.com/orgs/dart-lang/projects/5


kevmoo:
CC @nshahan

PR #562, opened 2017-04-21
chore(ng 3.0.0): deploy to default firebase project, etc

chalin:
Changes:
- deploy to `default` firebase project
- pubspec updates:
  - `angular2: ^3.0.0`
  - `angular_components: ^0.5.0`
  - `angular_test: ^1.0.0-beta+2`
- site var `acx_api` set to "/components/api".


chalin:
Waiting until 3.0.0 is released.

_config.yml
```diff
@@ -20,7 +20,7 @@ dart_vm:  "https://www.dartlang.org/dart-vm"
 flutter:  "https://flutter.io"
 dartino:  "https://dartino.org"
 dart_api: "https://api.dartlang.org/stable"
-acx_api: "https://www.dartdocs.org/documentation/angular_components/latest"
+acx_api: "/components/api"
```
kwalrath: We need to change this back, right?

kwalrath: Never mind, we don't. We're going to keep the canonical docs (a copy of the latest docs) at /components/api.

Issue #553, opened 2017-04-18
[ng] Add angular_test-based tests to all tutorial parts


chalin:
Contributes to #195

chalin:
Closing this for now given that tests have been written for all parts. `AppComponent` tests for toh-5 and toh-6 are pending resolution of https://github.com/dart-lang/angular_test/issues/61.

PR #565, opened 2017-04-22
[ng] toh-1 to toh-5 test cleanup

chalin:
Contributes to #553


kevmoo:
We should talk about finals in methods RE dart style guidelines. Cc @munificent

PR #569, opened 2017-04-25
[ng] chore: tweaks to cheatsheet build process

chalin:
Build script updates necessary to support these recent changes to the cheatsheet: https://github.com/dart-lang/angular2/pull/359.

The `EventEmitter` replacement will be rendered like this:
screen shot 2017-04-25 at 1 38 31 pm


chalin:
No, that was unintentional. Thanks for catching that. (I was trying out the older ng.io scripts so that is what might have created the .json file in the old location. I'll investigate.)

PR #576, opened 2017-04-26
[ng] template-syntax: add sizer component test

chalin:
The sizer component test is used to illustrate testing of `@Output` fields.


kwalrath:
A SWE should review this.

Issue #591, opened 2017-05-01
Don't depend on web-angular-simple template


kwalrath:
(I thought we already had a bug for this, but I can't find it.)

We have too many stagehand templates (https://github.com/google/stagehand/issues/390).

We need to have just one AngularDart template, so web-angular-simple has to go. We can add some/all of its features (e.g. test/) to web-angular. We can modify web-angular, but it needs to still feature material design and be simple yet interactive.

cc @chalin @filiph

filiph:
Yes, this closely relates to https://github.com/google/stagehand/issues/390. What are the next steps? Does someone need to rewrite the web-angular template over at stagehand first? Any takers?

chalin:
Here is a proposal:

## stagehand
- Drop `web-angular-simple`.
- `web-angular`
  - Drop the `<script>` dynamically setting `<base href>`.
  - Optional: add basic tests.
## webdev/angular docs
- [**Get started** page](https://webdev.dartlang.org/guides/get-started) stays as is, built off of `web-angular`.
- [**Setup** page](https://webdev.dartlang.org/angular/guide/setup) is changed to use [WS's integrated support](https://www.jetbrains.com/help/webstorm/2017.1/using-github-integration.html) for [**creating a project from a git repo**](https://www.jetbrains.com/help/webstorm/2017.1/setting-up-a-local-git-repository.html#cloneExistingRepository). We can either give the link to the [quickstart git](https://github.com/angular-examples/quickstart.git) directly in the page, or show users how to find it:
    > ![screen shot 2017-05-01 at 5 38 23 pm](https://cloud.githubusercontent.com/assets/4140793/25597329/22eb20e6-2e9c-11e7-96b2-5f749e4cdec0.png)

  If we show readers where to get the git URL from, then they'll be able to create projects from any one of the [Angular example repos](https://github.com/angular-examples/).

kwalrath:
I didn't know WS had that support! How easy is it to use? The less setup, the better.

chalin:
Almost as easy as using a template:

- Choose "**Checkout from Version Control > Git**" from either the **startup window**, or the **VCS** menu:
  > ![screen shot 2017-05-01 at 7 05 26 pm](https://cloud.githubusercontent.com/assets/4140793/25598184/90c78b90-2ea1-11e7-951f-3c9610bba7b5.png)

- In the dialog that pops up, enter the repo URL, parent directory and project name, then click Clone:
[...]

PR #595, opened 2017-05-02
[ng] quickstart updates

chalin:
This sync's up the `quickstart` example so that it more closely matches what was previously contained in the `web-angular-simple` stagehand template.

Contributes to #591, google/stagehand#390.


examples/ng/doc/quickstart/.docsync.json
```diff
@@ -1,3 +1,5 @@
 {
-  "title": "QuickStart"
+  "title": "Tutorial Starter App",
```
kwalrath: The page is titled "The Starter App", so having "Tutorial Starter App" as the link might be a bit confusing. How about just "Starter App"?

chalin: "Starter App" it is.

PR #597, opened 2017-05-03
[ng] chore: store example boilerplate files with other dart-doc-syncer assets

chalin:
Followup to #595.


chalin:
Here is the followup issue: #598

PR #603, opened 2017-05-06
Emulated view encapsulation with classes (instead of attributes)

sestegra:
With Angular 3, emulated view encapsulation is done with classes.


kwalrath:
Thanks very much for the PR, @sestegra. @chalin, could you take a look?

chalin:
I took a quick look this morning and it seemed good. I'll take a careful look shortly, thanks.

PR #631, opened 2017-05-18
[ng] update to 3.1.0

chalin:


examples/ng/api/common/pipes/pubspec.yaml
```diff
@@ -4,7 +4,7 @@ version: 0.0.1
 environment:
   sdk: '>=1.23.0 <2.0.0'
 dependencies:
-  angular2: ^3.0.0
+  angular2: ^3.1.0-beta+1
```
kwalrath: I'll think we'll want to just do ^3.1.0 (once it ships).

chalin: Right. We're just waiting for the pub release.

kwalrath:
Other than changing 3.1.0-beta+1 to 3.1.0, LGTM.

PR #634, opened 2017-05-18
Link to new codelabs

kwalrath:
Staged: https://kw-webdev-dartlang-1.firebaseapp.com/codelabs


kwalrath:
Merged because I think this shouldn't be too contentious (and we can always iterate).

chalin:
LGTM

PR #644, opened 2017-05-20
chore: gulp/example-frag - don't complain if symlink exists

chalin:
This fixes an issue with the `gulp/example-frag` tasks which report a warning if run repeatedly on a local repo. This fix simply aids from having the warning issued.

Also, misc angular.io file sync (mainly white-space changes).


chalin:
@kwalrath - minor updates, and travis is happy. I'll merge now.

PR #647, opened 2017-05-20
chore: gulp/ngio-get - tweak synced files

chalin:


chalin:
Minor script tweak. Merging.

PR #649, opened 2017-05-20
[ng] toh-3 updates + selector renames

### LINKED ISSUE #639: [ng] toh-3 sync needed to bring in the last tech writer edits
E.g., toh-3 still uses "we" vs "you".

chalin:
The main file to review is `toh-pt3.jade`.

Staged at: https://angulardart-org-dev.firebaseapp.com/angular/tutorial/toh-pt3

Fixes #639

This syncs with the latest angular.io version.
This also drops the `my-` prefix for some selectors where it is unnecessary.


src/angular/_jade/ts/patch/tutorial/toh-pt3.jade.patch
```diff
@@ -0,0 +1,352 @@
...
++  Just as you've done for the common Angular directives, you tell Angular
++  about the hero detail component by listing it in the metadata `directives`
++  list:
++
+++makeExcerpt('lib/app_component.dart', 'directives')
++
++:marked
++  Refresh the browser. It works!
++
++  ## App design changes
```
chalin: @kwalrath - I'm unsure about this heading text. On the TS side it's "What changed?" but I didn't find that appropriate here (since it could imply that the section would described what was changed in order to get the app to work ... as was done in the section just before this heading).

kwalrath: "App design changes" sounds reasonable to me.

PR #651, opened 2017-05-22
Add version info for AngularDart Components

kwalrath:
Also give a bit of information about the TypeScript version of Angular.

Staged at: 

https://kw-webdev-dartlang-1.firebaseapp.com/components
https://kw-webdev-dartlang-1.firebaseapp.com/angular/version

/cc @nshahan @matanlurey


chalin:
How about putting both versions next to the top heading of the versions page: something like
# AngularDart Versions <sub>3.0 / 0.5.1</sub>

nshahan:
Should this version number be something that we update every time we publish a new version? Is there any way to query for that info from pub?

kwalrath:
We have to update the example code every time we publish a new version. So we can't just grab the version number from pub, we have to confirm that everything builds and republish the site.

kwalrath:
@chalin interesting idea... maybe so. I'll file an issue.

nshahan:
Ok. Looks good. Thanks for the update!

PR #638, opened 2017-05-19
chore: kramdown plugin extension supporting code-excerpts

chalin:
Staged at: https://angulardart-org-dev.firebaseapp.com/playground

Contributes to
- #637 - in helping us transition from Jade to markdown
- #374 - [ng] fix makeExample stylePatterns

Among other variants, it supports markdown like:

    <?code-excerpt "quickstart/lib/app_component.dart" title?>
    ```
      import 'package:angular2/angular2.dart';

      [!@Component(
        selector: 'my-app',
        template: '<h1>Hello {{name}}</h1>')!]
      class AppComponent {
        var [!name!] = 'Angular';
      }
    ```
And renders it like this:
> screen shot 2017-05-19 at 4 54 07 pm

Issue #637, opened 2017-05-19
decrease technical debt of site-www and site-webdev


chalin:
As @filiph, mentioned, this is in part motivated by the currently poor authoring experience (#208) -- i.e., it takes too long for a site/page refresh after a minor edit:

Proposals:
- [x] Transition away from Jade towards more use of markdown
  - [x] tutorial index, toh-0
  - [x] toh-1 #691 
  - [x] toh-2 #721
  - [x] toh-3 #722
  - [x] many other pages were handled via #713
  - [x] all remaining files #768 
- [x] Curate repo, removing obsolete Jade and Harp support files #896
- [ ] Share site infrastructure (Jekyll plugins, build scripts, etc.)
  - [x] Setup a site-shared submodule
  - [x] Factor out code excerpter. Done! Currently at https://github.com/chalin/code-excerpter. Also: #1402

cc @filiph @kwalrath

chalin:
Closed via #1472

Issue #374, opened 2017-02-15
support code excerpt highlighting


chalin:
**[ng] fix makeExample stylePatterns**

I though that this was working at some point, but it isn't now. I've spent some time investigating, but it is difficult to isolate the issue. Strangely, the same feature works under ng.io

chalin:
We emulate Harp `partial` processing ourselves, and the problem is that the partials for code fragments haven't been substituted by the time that the style patterns are applied.

chalin:
Consider also adapting the oneway translation, and applying `makeEx*` path renames during file synchronization.

PR #653, opened 2017-05-23
chore(Jekyll Jade plugin): support <?code-excerpt?>

chalin:
This adds support for <?code-excerpt?> processing instructions found in Jade markdown.

Staged at: https://angulardart-org-dev.firebaseapp.com/playground/jade-test

Factored out code-excerpt processing methods from the NgMarkdown plugin, placing it into a mixin. Both NgMarkdown and the Jade plugins use the mixin.

Followup to #638, and so contributes to
- #637 - in helping us transition from Jade to markdown
- #374 - [ng] fix makeExample stylePatterns


src/playground/jade-test.jade
```diff
@@ -11,7 +11,7 @@ permalink: /playground/jade-test
   Title and code not tied to a file:
```
kwalrath: Delete the "(Code excerpts...)" paragraph?

src/playground/jade-test.jade
```diff
@@ -63,3 +63,14 @@ permalink: /playground/jade-test
   ```
   <my-app>Loading...</my-app>
   ```
+  ## Converted (manually)
```
kwalrath: This section doesn't exactly match what's in the non-Jade version. Which one should change?

src/playground/jade-test.jade
chalin: done

src/playground/jade-test.jade
chalin: Thanks. I've made them consistent.

chalin:
> I see that ** is treated as bold (in the transformers section of the pubspec). Can you fix that now?

Good catch!  Looking into it.

chalin:
Fixed and redeployed to the staging server.

PR #654, opened 2017-05-24
chore: use latest firebase-tools, specify project on serve

chalin:
The `check-links.sh` script needs to serve the locally built site. Firebase now requires authentication to serve locally. Authentication requires that the repo have a `FIREBASE_TOKEN` environment variable defined in the Travis settings of `site-webdev`.

The `check-links.sh` script will warn if `FIREBASE_TOKEN` isn't defined, but it won't fail. This way, we don't force anyone to define a token in Travis (esp. since doing so is moot for folks who don't have permissions over the `dev` firebase project).

cc @Sfshaza


kwalrath:
Thanks for tracking this down and working around it, Patrice!

chalin:
Cf. https://github.com/firebase/firebase-tools/issues/302

PR #662, opened 2017-05-26
chore(code-excerpt PI): support class attr, and PI w/o any args

chalin:
Also included other misc. changes needed for the testing branch.


_config.yml
```diff
@@ -23,6 +23,7 @@ dart_api: "https://api.dartlang.org/stable"
 acx_api: "/components/api"
 flutter_api: "http://docs.flutter.io"
 dartino_api: "https://dartino.github.io/api"
+api: "https://www.dartdocs.org/documentation"
```
kwalrath: "api" seems very generic. Maybe "package_api"?

kwalrath: Or ddo_api? (It's not obvious what it means unless you already know what it means... but maybe that's OK.)

chalin: I was looking for a short name because URLs into those API pages tend to be quite long.
Aside from core Dart libraries, dartdocs.org is the default API repo, so `api` seemed ok.

PR #664, opened 2017-05-26
[ng] toh/index update + new CSS styles for file trees

chalin:
Note that the new toh/index file is in markdown, not Jade :)

Staged at https://angulardart-org-dev.firebaseapp.com/angular/tutorial/


src/playground/html-test.html
```diff
@@ -0,0 +1,16 @@
...
+    </div>
+    <div class="file">web</div>
+    <div class="children">
+      <div class="file">index.html</div>
+      <div class="file">main.dart</div>
+      <div class="file">styles.css</div>
+    </div>
+    <div class="file">pubspec.yaml</div>
+  </div>
+</div>
```
kwalrath: The rendered version of this looks a bit sad: https://angulardart-org-dev.firebaseapp.com/playground/html-test

(I don't think that matters for this PR, though.)

PR #672, opened 2017-06-01
chore(travis): upgrade to ruby 2.4

chalin:


chalin:
For the record, this PR actually updated from Ruby 2.3.0 to 2.3.3. Some of the gems we use haven't been updated to 2.4 yet, so we'll wait a bit before switching to 2.4.

PR #675, opened 2017-06-01
dartdevc page

kwalrath:
Staged at:

https://kw-webdev-dartlang-1.firebaseapp.com/tools
https://kw-webdev-dartlang-1.firebaseapp.com/tools/dartdevc


src/tools/dartdevc.md
```diff
@@ -0,0 +1,70 @@
...
+lets you run and debug your Dart web app
+in any modern browser.
+As of SDK release 1.24,
+`pub build` and `pub serve` support dartdevc.
+
+Unlike the original Dart-to-JavaScript compiler ([dart2js](/tools/dart2js)),
+dartdevc is modular,
+which allows `pub serve` to do incremental rebuilds.
+You can edit your Dart files,
+refresh in Chrome (or another modern browser),
```
chalin: Rather than
> refresh in Chrome (or another modern browser),

why not just
> refresh [the/your] browser

src/tools/dartdevc.md
```diff
@@ -0,0 +1,70 @@
...
+and see your edits almost immediately.
+This speed is possible because pub compiles only _your_ package,
+not all the packages that you depend on.
```
chalin: Maybe:
> ... because pub compiles only updated packages,
> not all packages that your app depends on.

src/tools/dartdevc.md
```diff
@@ -0,0 +1,70 @@
...
+
+To compile with dartdevc, your web app's code—and
+all packages you depend on—**must be type safe**,
+as described in [Sound Dart](https://www.dartlang.org/guides/language/sound-dart).
+
+
+## Building with dartdevc
+
+To opt into using dartdevc, do one of the following:
```
chalin: Maybe:
> To use the dartdevc, ...

src/tools/dartdevc.md
```diff
@@ -0,0 +1,70 @@
...
+
+  ```yaml
+  web:
+    compiler:
+      debug: dartdevc
+  ```
+
+## Testing with dartdevc
+
+You can use dartdevc to run your tests in Chrome much more
```
chalin: Only in Chrome?

src/tools/dartdevc.md
```diff
@@ -0,0 +1,70 @@
...
+  
+* Modify your pubspec. A new `web` key supports a
+  single key named `compiler` that maps mode names to compilers.
+  For example, to default to dartdevc in debug mode,
+  put the following in your `pubspec.yaml` file:
+
+  ```yaml
+  web:
+    compiler:
+      debug: dartdevc
```
chalin: Possibly mention that `debug` is the name of pub's [default build mode](https://webdev.dartlang.org/tools/pub/pub-build#options).

jakemac53:
Note that the `pub serve` and `pub build` pages also should be updated with the options for enabling dartdevc

kwalrath:
> Also, maybe mention if there is any possible interplay/interference between the use of the ddc and the dart_to_js_script_rewriter transformer.

I seem to remember hearing that the dart_to_js_script_rewriter transformer runs only in release mode. If that's true, then if you specify dartdevc only for debug mode (which is what I believe we should recommend), you won't run into issues.

Does that sound right, @jakemac53?

src/tools/dartdevc.md
kwalrath: We have a Chrome extension coming out that'll make debugging Dart in Chrome easier, so I think we want to recommend Chrome. Even if all browser debuggers were equal, I suspect that since we tend to use Chrome internally, you're likely have the best experience using Chrome. 

@jakemac53, what do you think?

kwalrath:
> Note that the pub serve and pub build pages also should be updated with the options for enabling dartdevc

Yep, I'm working on those, perhaps for a separate PR. I should also take a look at any place that refers to Dartium or dart2js.

src/tools/dartdevc.md
jakemac53: I don't have much of an opinion here, depending on how your app is set up if you load in dartium it might actually load the dart code (not dartdevc code).

src/tools/dartdevc.md
jakemac53: Chrome is the only _supported_ browser for now, but it does work in many cases for other browsers.

jakemac53:
In terms of the dart_to_js_script_rewriter, it should work exactly the same as it does with dart2js. Same goes for the `packages/browser/dart.js` script. Dartdevc creates a single file by the same name that dart2js would so it is a drop in replacement.

src/tools/dartdevc.md
kwalrath: Actually, pub's default mode is `release` for `pub build` and `debug` for `pub serve` (according to my hazy memory, which I just checked against the docs for [pub build](https://webdev.dartlang.org/tools/pub/pub-build#options) & [pub serve](https://webdev.dartlang.org/tools/pub/pub-serve#options)).

kwalrath:
PTAL, and especially look for "PENDING" questions.

src/tools/dartdevc.md
kwalrath: Done.

src/tools/dartdevc.md
kwalrath: Did something similar...

src/tools/dartdevc.md
kwalrath: But I did add a mention of what the mode means.

src/tools/dartdevc.md
```diff
@@ -0,0 +1,93 @@
...
+refresh in Chrome (or another modern browser),
+and see your edits almost immediately.
+This speed is possible because pub compiles only updated packages,
+not all packages that your app depends on.
+
+## Prerequisites
+
+To compile with dartdevc, your web app's code—and
+all packages you depend on—**must be type safe**,
+as described in [Sound Dart](https://www.dartlang.org/guides/language/sound-dart).
```
chalin: Developers unfamiliar with the notion of soundness might feel overwhelmed at the prospect of reading the 15 page article on Dart soundness :).

In practice, all they have to do is ensure that their **code compiles w/o errors**, but this is true independent of type safety. So it seems a bit strange to have as a _prerequisite_ that essentially reads "To compile with dartdevc, your web app's code must be free of (dartdevc) compiler errors" :). 

Rather than mention this as a prerequisite, maybe we can just have a note that the dartdevc has stricter typing rules?

src/tools/dartdevc.md
```diff
@@ -0,0 +1,93 @@
...
+
+For best performance, **put implementation files under `/lib/src`**,
```
chalin: We're talking about performance optimization here ("For best performance ..."), so this isn't really a _prerequisite_ for use. Maybe we should drop the Prerequisite heading in favor of a Performance section and a separate note about the stricter typing rules of the ddc (see previous comment for some context)?

src/tools/dartdevc.md
```diff
@@ -0,0 +1,93 @@
...
+
+The Dart development compiler _(dartdevc_, also known as _DDC)_
+lets you run and debug your Dart web app
+in any modern browser.
+As of SDK release 1.24,
+`pub build` and `pub serve` support dartdevc.
+
+Unlike the original Dart-to-JavaScript compiler ([dart2js](/tools/dart2js)),
+dartdevc is modular,
+which allows `pub serve` to do incremental rebuilds.
```
chalin: > **dartdevc is modular**, which allows `pub serve` to do incremental rebuilds.

Using _modular_ in this way is misleading &mdash; it seems to be qualifying the ddc itself as having been written in a modular way. Instead:

> Unlike the original Dart-to-JavaScript compiler ([dart2js](/tools/dart2js)),
> the dartdevc supports _incremental_ compilation.

Actually, a questions for @jakemac53: does it support incremental compilation or only incremental _re-_ compilation?

chalin: (Maybe we want to mention early that the unit of the incremental (re-)compilation is a "module"?)

src/tools/dartdevc.md
```diff
@@ -0,0 +1,93 @@
...
+
+## Prerequisites
+
+To compile with dartdevc, your web app's code—and
+all packages you depend on—**must be type safe**,
+as described in [Sound Dart](https://www.dartlang.org/guides/language/sound-dart).
+
+For best performance, **put implementation files under `/lib/src`**,
+instead of anywhere else under `/lib`.
+Also, **don't import from `lib/src`**.
```
chalin: I'm feeling that this paragraph needs to say more to explain why/how these do's and don't's are impacting performance. We may need to go into some detail about how **module** boundaries are determined. (Ah, it is interesting that there is no mention yet of modules in this page.)

Maybe start by mentioning that the performance of the incremental compiler will, of course, depend on how the app is modularized. If the entire app finds itself in a single module, then the ddc will perform no better than dart2js. Then we should list the rules by which module boundaries are established. That way developers can decide how to reorganize their code, if necessary.

src/tools/dartdevc.md
chalin: Here's another possibility:
> Unlike the original Dart-to-JavaScript compiler ([dart2js](/tools/dart2js)),
> the dartdevc is an _incremental_ compiler that emits modular JavaScript.

kwalrath: + @vsmenon since I believe @jakemac53 is OOO.

src/tools/dartdevc.md
kwalrath: I could just point them to https://www.dartlang.org/guides/language/sound-problems, which is a very practical guide to converting your code.

chalin: SGTM

src/tools/dartdevc.md
kwalrath: I want to cover this in a separate page. This page should have only the most essential information for using dartdevc.

src/tools/dartdevc.md
```diff
@@ -0,0 +1,93 @@
...
+{% comment %}
+TODO: add link to FAQ once it's up.
+{% endcomment %}
+
+
+## Building with dartdevc
+
+To build using dartdevc, do one of the following:
+
+* Modify your pubspec. A new `web` key supports a
```
kwalrath: Done.

src/tools/dartdevc.md
kwalrath: I'm changing it to "Preparing your code"

src/tools/dartdevc.md
vsmenon: Funny enough, I was just starting to update the DDC readme.  Staging that here:

https://github.com/dart-lang/sdk/blob/ddc-readme/pkg/dev_compiler/README.md

I tend to use *modular* instead of *incremental* as DDC compiles at the granularity of modules (on initial compile or recompilation).  This supports incremental compilation at an application level as the build system (e.g., pub or bazel) only re-invokes DDC for modules that need to be regenerated.

src/tools/dartdevc.md
vsmenon: One thing we could expand on is runtime type errors.  When someone runs an existing app on DDC for the first time, there is a good chance they will encounter runtime type errors - which will need to be fixed just like the static ones.

src/tools/dartdevc.md
```diff
@@ -0,0 +1,108 @@
...
+**[PENDING: that's true, right? Would it be ok to do `import 'package:mypackage/src/myfile.dart'`?]**
+See the
+[pub package layout conventions](https://www.dartlang.org/tools/pub/package-layout)
+for more information about how to structure your code.
+{% comment %}
+TODO: add link to FAQ once it's up.
+{% endcomment %}
+
+
+## Building with dartdevc
```
chalin: Maybe:
> ## Configuring pub to use dartdevc

The original heading mentions "building" but we cover building and serving.

src/tools/dartdevc.md
```diff
@@ -0,0 +1,108 @@
...
+
+To build using dartdevc, do one of the following:
```
chalin: Maybe
> To have pub use dartdevc, do one of the following:

src/tools/dartdevc.md
```diff
@@ -0,0 +1,108 @@
...
+
+* Modify your pubspec. A new top-level key named `web` supports a
+  single key named `compiler` that maps mode names to compilers.
```
chalin: How about:
> ... `compiler` that maps mode names (`debug` or `release`) to compilers.

Though I'm thinking that I prefer:
> ... `compiler` that allows you to name the compiler that pub should use for the build/serve modes `debug` and `release`.

src/tools/dartdevc.md
```diff
@@ -0,0 +1,108 @@
...
+
+* Use the new `--web-compiler` flag to `pub build` or `pub serve`.
+  Specify the `dartdevc` option:
+
+  ```
+  pub serve --web-compiler=dartdevc
+  ```
+
+<aside class="alert alert-info" markdown="1">
+**Note:**
```
chalin: Should this be "Notes:" with the two statements that follow each be a list item in a bulleted list? (I'm suggesting this because the two statements aren't related.)

src/tools/dartdevc.md
kwalrath: The testing section uses pub too...

I actually think of pub serve as building, too. It's actually the primary way you'll build using dartdevc.

src/tools/dartdevc.md
kwalrath: According to the [pub serve docs](https://webdev.dartlang.org/tools/pub/pub-serve#options), 'debug' and 'release' aren't the only possible modes. So... I've modified the wording, but not exactly like this.

src/tools/dartdevc.md
kwalrath: Nah. :) But I've deleted the second sentence, since I wasn't absolutely sure about it.

src/tools/dartdevc.md
```diff
@@ -0,0 +1,108 @@
...
+---
+
+The Dart development compiler _(dartdevc_, also known as _DDC)_
+lets you run and debug your Dart web app
+in any modern browser.
+As of SDK release 1.24,
+`pub build` and `pub serve` support dartdevc.
+
+Unlike the original Dart-to-JavaScript compiler ([dart2js](/tools/dart2js)),
+dartdevc is an incremental compiler that emits modular JavaScript.
```
kwalrath: Done (but slightly differently).

src/tools/dartdevc.md
```diff
@@ -0,0 +1,108 @@
...
+When you use `pub serve` with dartdevc,
+you can edit your Dart files,
+refresh in Chrome (or another modern browser),
+and see your edits almost immediately.
+This speed is possible because pub compiles only updated packages,
+not all the packages that your app depends on.
+
+<aside class="alert alert-info" markdown="1">
+**Note:**
+The first compilation with dartdevc takes longer than it would with dart2js.
```
kwalrath: Done (more or less).

src/tools/dartdevc.md
```diff
@@ -0,0 +1,108 @@
...
+To compile with dartdevc, your web app's code—and
+all packages you depend on—must be **type safe**.
+For details on how to make your code type safe, see
+[Sound Dart: Fixing Common Problems.](https://www.dartlang.org/guides/language/sound-problems)
+
+<aside class="alert alert-info" markdown="1">
+**Tip:**
+Your code might already be type safe.
+You can check it just by building with dartdevc, or you can use
+[`dartanalyzer --strong`](https://www.dartlang.org/guides/language/sound-problems#command-line-analyzer).
```
kwalrath: Done.

src/tools/dartdevc.md
kwalrath: Done.

kwalrath:
OK to publish?

ptal, @chalin & @vsmenon.

chalin:
Has the staging server been updated?

src/tools/dartdevc.md
```diff
@@ -0,0 +1,121 @@
...
+{% endcomment %}
+
+
+## Building with dartdevc
+
+To make pub use dartdevc, do one of the following:
+
+* **Modify your pubspec.** A new top-level key named `web` supports a
+  single key named `compiler` that maps mode names
+  (such as `debug` and `release`) to compilers (`dartdevc`, `dart2js`, or `none`).
```
chalin: What happens when `none` is used?

kwalrath: Native Dart, no JavaScript, I believe. I should clarify that somewhere... later.

PR #679, opened 2017-06-02
[ng] upgrade to ACX 0.5.2 where necessary

### LINKED ISSUE #674: Update angular_components API docs to 0.5.2
/cc @nshahan

chalin:
Fixes #674


_config.yml
```diff
@@ -87,8 +87,8 @@ custom:
     dev-version:
     # url-next-vers: https://webdev-angular3-dartlang-org.firebaseapp.com
   components:
-    stable-version: 0.5.1
-    stable-version-full: 0.5.1
+    stable-version: 0.5.2
```
chalin: Kathy, do we want `stable-version` to be just `0.5`?
(My guess is that you wanted "2 digits of precision" :). Pls confirm.)

kwalrath: Yes, 2 digits of precision is what we want, I think. So `0.5.2`. (On https://webdev.dartlang.org/angular/version we promised to reflect the latest stable release.)

Note that this is independent of the version number we specify in Stagehand and the samples, which is 1 digit of precision unless we're relying on something introduced after the major release. (See https://github.com/google/stagehand/issues/419.)

/cc @matanlurey and @nshahan

kwalrath: Ah, I see that you updated the samples, too. I'm not sure what we should do. It seems fine to update these here, but updating every other occurrence throughout our repos is a pain (see https://github.com/google/stagehand/issues/419#issuecomment-305577363).

On the other hand... Since these examples don't use Stagehand, maybe it's fine to have them specify a higher version.

PR #689, opened 2017-06-05
plugins/code_excerpt_processor: strip file vers suffix from title

chalin:


chalin:
No content changes, only plugin updates. Merging.

PR #690, opened 2017-06-05
gulp/jade-to-md.js - Jade to Markdown converter helper

chalin:
The converter only operates on the file content, it doesn't rename the file (this makes it easier to view diffs). The code_excerpt_updater must also be run separately.


chalin:
No content changes, only new gulp task. Merging.

PR #718, opened 2017-06-08
[ng] drop toh-0 sources from repo and generate from quickstart instead

### LINKED ISSUE #642: [ng] consolidate quickstart and toh-0 examples
We ended up having two examples because one was automatically derived from the stagehand template `web-angular-simple`. Now that that template has been dropped, consider consolidating these by either dropping one, or at least deriving one from the other.

chalin:
The QuickStart and toh-0 examples are almost identical. This PR consolidates the sources:
- Add `docregion`s to the QS sources.
- Delete the toh-0 sources.
- Add tasks to generate toh-0 sources (in the local tmp folder). This allows us to continue to refer to toh-0 source fragments in the prose (which we do in the tutorial Starter page and the soon-to-be-published component test page).

There is no **changes to the docs** (other than a fix to the display of pubspec.yaml, which only showed part of the file in the makeTabs, but should have shown all of the file).

Fixes #642.


kwalrath:
Uh oh, build failed.

chalin:
This should be fixed now. Let's see if Travis agrees.

PR #717, opened 2017-06-08
[ng] toh-0: clarify that the live example is quickstart

chalin:
The new prose renders like this:
> screen shot 2017-06-08 at 8 53 13 am

PR #720, opened 2017-06-09
[ng] app setup customization, and toh-0 live example tweak

chalin:
Updates:
- Setup: add minor project customization steps
- Toh-0: clarify that the live example is quickstart

Supersedes #717

Staged at:
- https://angulardart-org-dev.firebaseapp.com/angular/guide/setup#customize-the-project
- https://angulardart-org-dev.firebaseapp.com/angular/tutorial/toh-pt0


src/angular/guide/setup.jade
```diff
@@ -80,6 +80,20 @@ a#get
...
+:marked
+  ## Customize the project
+
+  Using WebStorm, or your favorite editor:
+
+  1. Open `pubspec.yaml`, and replace the top two entries:
+    * **name:** `angular_tour_of_heroes` _(or any other
+     [valid package name](https://www.dartlang.org/tools/pub/pubspec#name))_
+    * **description:** Tour of Heroes _(or a description suitable for your project)_
+  1. Open `web/index.html`, and replace the text of the `**<title>**` element with a
```
kwalrath: Move those `s inside the **s.

kwalrath: Also... we should update the screenshot to match the customizations, right?

src/angular/guide/setup.jade
```diff
@@ -80,6 +80,20 @@ a#get
   in a terminal window, go to the project root and run `pub get`.
 
 
+:marked
+  ## Customize the project
+
+  Using WebStorm, or your favorite editor:
+
+  1. Open `pubspec.yaml`, and replace the top two entries:
+    * **name:** `angular_tour_of_heroes` _(or any other
```
kwalrath: The name should already be set correctly for you if you've followed the WebStorm instructions above. Not sure how to phrase this.

chalin: No, when you get a project from a git repo, WS doesn't update the name in the pubspec. I just checked.

src/angular/guide/setup.jade
chalin: Oups. Missed that! Thanks.

chalin: Fixed

chalin:
This has turned out to be a bit more complicated than I thought, but it has been a worth-while exercise IMHO.

In summary, when WebStorm creates a project from a git repo, it doesn't rename the project to match the chosen directory name. So, I've renamed the quickstart project from `angular_quickstart` to the more generic `angular_app` (so that it is a more reasonable default to keep for any given project). I've none-the-less adapted the setup instructions to explain how a rename can be done (via a global search-and-replace). So the setup "Customize the project" step now reads as follows:

screen shot 2017-06-09 at 3 19 19 pm

@kwalrath PTAL

kwalrath:
SGTM. Is it really true that you'll need to do a pub get in WebStorm but not if you use the command line or a different editor to make the changes?

chalin:
It _is_ necessary in WS to click "Get dependencies", otherwise WS won't pickup the project name change.

I don't know about other editors. If one is working from the **command line**, then any subsequent invocation of, say, `pub` or the `dartanalyzer` will naturally pickup the latest pubspec content.

So, is this approved as is?

kwalrath:
Here's a new screenshot to use: 
image

src/angular/guide/setup.jade
```diff
@@ -80,6 +80,30 @@ a#get
...
+  ## Customize the project
+
+  Using WebStorm, or your favorite editor:
+
+  1. Open **`web/index.html`**, and replace the text of the **`<title>`** element
+     with a title suitable for your app. For example: `<title>`_`Angular Tour
+     of Heroes`_`</title>`.
+
+  1. Open **`pubspec.yaml`**, and update the **description** to suit your project.
+     For example: `description: `_` Tour of Heroes`_.
```
kwalrath: Let's just use `description: Tour of Heroes`

Otherwise, it looks like there's no space after the :, and also there's no need to italicize the example (because it's an actual example, not stand-in text describing the example).

chalin: Fixed.
Also removed the italics from the `<title>`.

chalin:
Added the new image and removed the statement that "Get dependencies" is necessary in WS (since it actually won't be necessary if the global search-and-replace is done.)

PR #724, opened 2017-06-11
Update URL to `portmanteaux.json` file

### LINKED ISSUE #8: samples files not found
_From @foundatron on February 7, 2016 18:7_

Sample application fails because the sample file is not found

Example: https://www.dartlang.org/docs/tutorials/fetchdata/#making-a-get-request
File not found: https://www.dartlang.org/samples-files/portmanteaux.json

```
Error: Not Found

The requested URL /samples-files/portmanteaux.json was not found on this server.
```

_Copied from original issue: dart-lang/www.dartlang.org#1549_

chalin:
- [x] Awaiting for update to https://gist.github.com/Sfshaza/83f7779d18a8bbe8ccb1.

I removed the copy of the code that was inlined in comments so as to keep things [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself).

Fixes #8


kwalrath:
The inlined comments were something we were doing to keep it easier to see what we were talking about. But maybe they're just a crutch we don't need any more...

chalin:
This is now ready for review. Thanks @Sfshaza for the update to the Gist!

chalin:
:)

Issue #713, opened 2017-06-07
[ng] Reorg code of examples in support of better pub/ddc modularization


chalin:
Contributes to #701.

Examples under `examples/ng/doc`
 - [x] architecture #726
 - [x] attribute-directives #714
 - [x] component-styles #714
 - [x] dependency-injection #727
 - [x] displaying-data + .jade --> .md #728
 - [x] forms + .jade --> .md #730
 - [x] hierarchical-dependency-injection + .jade --> .md #731
 - [x] lifecycle-hooks + .jade --> .md #732
 - [x] pipes + .jade --> .md #734
 - quickstart: no src reorg needed
 - [x] router + .jade --> .md #736
 - [x] security #727
 - [x] server-communication + .jade --> .md #753
 - [x] structural-directives + .jade --> .md #747
 - [x] template-syntax + .jade --> .md #748
 - toh-0: derived from quickstart
 - toh-1: no src reorg needed
 - toh-2: no src reorg needed
 - [x] toh-3 #744
 - [x] toh-4 + .jade --> .md #745
 - [x] toh-5 + .jade --> .md #750
 - [x] toh-6 + .jade --> .md #751
 - [x] user-input + .jade --> .md #746

Codelabs:
- [x] pirate badge (ng2) codelab #755
- [x] acx codelab #759

chalin:
All done! 😅  🎉

PR #727, opened 2017-06-11
[ng] src reorg of dependency-injection example

chalin:
Contributes to #713


src/angular/_jade/ts/_cache/guide/dependency-injection.jade
```diff
@@ -25,7 +25,7 @@ block includes
 
   Let's start with the following code.
 
-+makeExample('dependency-injection/ts/app/car/car-no-di.ts', 'car', 'app/car/car.ts (without DI)')
++makeExcerpt('app/src/car/car-no-di.ts (without DI)', 'car')
```
kwalrath: The title has changed. The visible filename should be .../car.dart.

chalin: I'm not a fan of giving paths to files that don't exist. (The exception is when we have a sequence of revisions of the same file, but then we label revisions with numeric suffixes, like `hero_1.dart`. But in the case we have here, these are two distinct source files. I'd rather give the path to the actual file so that an interested reader can find it.)  Do you agree?

kwalrath: I'd rather be consistent with the Angular TS docs. These extra files are just a doc implementation detail. I don't expect people to be cross checking against the source repo (and if they are, they'll quickly enough figure out what's going on).

kwalrath: Another option is just to name the class, not the file.

kwalrath:
I don't see any mention of `src` in this page...

src/angular/_jade/ts/_cache/guide/dependency-injection.jade
chalin: Fixed

chalin:
> I don't see any mention of src in this page...

I'm not sure what you mean, but I did review the page, looking for places that might need to mention `src` and I did find one (along with a mis-named file) -- see the [latest commit](https://github.com/dart-lang/site-webdev/pull/727/commits/1275bc2d33250bc1fa88cfc41eb2adb92655682d).

chalin:
Travis is happy, and you approved, so I'll go ahead and merge this PR.

PR #730, opened 2017-06-12
[ng] src reorg of forms example, and convert Jade to .md

chalin:
I've cross checked the original to the markdown-generated version of the page (just as I did for the previous pages too).

Contributes to #713


chalin:
I'll deal with the scrollbar issue separately. Merging.

PR #732, opened 2017-06-12
[ng] src reorg of lifecycle-hooks example, and convert Jade to .md

chalin:
I've cross checked the original to the markdown-generated version of the page

Contributes to #713


src/angular/guide/lifecycle-hooks.md
```diff
@@ -0,0 +1,724 @@
...
+Clearly our implementation must be very lightweight or the user experience will suffer.
+
+## AfterView
+
+The *AfterView* sample explores the `AfterViewInit` and `AfterViewChecked` hooks that Angular calls
+*after* it creates a component's child views.
+
+Here's a child view that displays a hero's name in an input box:
+
+<?code-excerpt "lib/src/after_view_component.dart (child-view)" title?>
```
kwalrath: The code says `my-child-view`, not `child-view`. Perhaps the code should change to `child-view`?

chalin: No, `child-view` was only the `docregion` tag name. I've changed the excerpt text to "child view" (which names what is excerpt, rather than suggest that it might be the selector name).

chalin:
Added the `src/` to the import paths of `app_component.dart`.

src/angular/guide/lifecycle-hooks.md
kwalrath: I thought they were changing the examples to not have `my-` when the rest of the component name already had a `-`. Not a big deal, either way.

PR #744, opened 2017-06-14
[ng] src reorg of toh-3 example; prose introduces `lib/src`

chalin:
Contributes to #713


src/angular/tutorial/toh-pt3.md
```diff
@@ -57,11 +57,13 @@ This file will hold the new `HeroDetailComponent`.
     case_](/angular/glossary#snake_case), each word separated by
     underscores, and end in `_component.dart`.  The `HeroDetailComponent`
```
kwalrath: end -> ends or ending

chalin: Really? If we remove the "each word ...", it currently reads:
> The component _file_ name should be spelled in [_lower underscore case_](/angular/glossary#snake_case), ..., and end in `_component.dart`.

This seems fine to me.

kwalrath: Yeah, you're right, it's fine grammatically, if a little hard to parse. It's a bit redundant, too, and takes you to a definition (snake_case) that isn't mentioned here. Maybe:

The component _file_ name should be in [snake case](/angular/glossary#snake_case)—lowercase with underscore separation—and end in `_component.dart`.

chalin: Yes, that is better. Fixed as you suggested.

src/angular/tutorial/toh-pt3.md
```diff
@@ -49,14 +49,13 @@ This file will hold the new `HeroDetailComponent`.
 <div class="l-sub-section" markdown="1">
   **Angular conventions**:
 
-  * The component _class_ name should be written in _upper camel case_ and
+  * The component _class_ name should be written in [upper camel case](/angular/glossary#pascalcase) and
```
kwalrath: good catch!

PR #769, opened 2017-06-22
Renaming Sound Dart pages to Strong Mode Dart.

Sfshaza:
This site has no links to the optional type articles and only this one reference to a renamed page.

@kwalrath


Sfshaza:
Done. I only found one place that needed fixing. (Same as the one you referenced.)

Issue #768, opened 2017-06-21
[ng] convert remaining .jade files to .md


chalin:
Contributes to #637.
 - [x] angular/api/index.jade
 - [x] angular/glossary.jade
 - [x] angular/guide/animations.jade
 - [x] angular/guide/architecture.jade
 - [x] angular/guide/attribute-directives.jade
 - [x] angular/guide/component-styles.jade
 - [x] angular/guide/dependency-injection.jade
 - [x] angular/guide/index.jade
 - [x] angular/guide/router/_milestone-nav.jade
 - [x] angular/guide/security.jade
 - [x] angular/guide/setup.jade
 - [x] angular/guide/testing.jade #793
 - [x] angular/tutorial/keep-app-running.jade
 - [x] playground/jade-test.jade

Also:
- Clean out the remaining `_data.json` files --> moved to #896

PR #785, opened 2017-06-26
[ng] architecture page in markdown

chalin:
This took longer to convert because of all the custom styling applied to each image. I tried to make things a little more uniform by using new classes.

The page content is unchanged except for the "alphabetical list of other important Angular features" in the "Wrap up" section which was converted into an actual list (rather than formatted as quoted text), and the prose was tweaked a bit.

Contributes to #768

Staged at: https://angulardart-org-dev.firebaseapp.com/angular/guide/architecture


src/angular/guide/architecture.md
```diff
@@ -0,0 +1,592 @@
...
+
+### Angular libraries
+
+<img class="image-left" src="/resources/images/devguide/architecture/library-module.png" alt="Libraries" width="200px">
+
+Angular ships as a collection of libraries within the
+[**angular2**](https://pub.dartlang.org/packages/angular2) package.
+
+Each Angular library name begins with the `angular2` prefix.
+For example, the main Angular library is `angular2/angular2`,
```
kwalrath: Now that I'm rereading this paragraph, it seems questionable. (Nothing you changed... I probably created this page.) For example, once you get into the API docs, you see names like angular2.router and angular2.platform.common, but never angular2.angular2. If a library name is even defined, it seems to be irrelevant.

Maybe:

Angular ships as a collection of libraries within the angular2 package. The main Angular library is angular2/angular2.dart, which most application modules import as follows: 

...

The angular2 package has other important libraries, such as ...

(It'd be nice to link to the API docs. It'd also be nice if the API docs were clearer about what you have to import to get access to that library... I'm not sure how to fix that.)

src/angular/guide/architecture.md
```diff
@@ -0,0 +1,592 @@
...
+## Components
+
+<img class="image-left" src="/resources/images/devguide/architecture/hero-component.png" alt="Component" width="200px">
+
+<div class="component-section-fix-overflow" markdown="1">
+  A _component_ controls a patch of screen called a *view*.
+
+  For example, the following views are controlled by components:
+
+  * The app root with the navigation links.
```
kwalrath: This list's formatting looks much better than before!

src/angular/guide/architecture.md
```diff
@@ -0,0 +1,592 @@
...
+* [Services](#services)
+* [Dependency injection](#dependency-injection)
+
+That's a foundation for everything else in an Angular application,
+and it's more than enough to get going.
+But it doesn't include everything you need to know.
+
+Here is a brief, alphabetical list of other important Angular features and services.
+Most of them are covered in this documentation (or soon will be).
+
```
kwalrath: The following list looks much better now!

src/angular/guide/architecture.md
```diff
@@ -0,0 +1,592 @@
...
+* [Directives](#directives)
+* [Services](#services)
+* [Dependency injection](#dependency-injection)
+
+That's a foundation for everything else in an Angular application,
+and it's more than enough to get going.
+But it doesn't include everything you need to know.
+
+Here is a brief, alphabetical list of other important Angular features and services.
+Most of them are covered in this documentation (or soon will be).
```
kwalrath: How about just deleting this sentence?

src/angular/guide/architecture.md
```diff
@@ -0,0 +1,592 @@
...
+
+- **Change detection**: The change detection documentation will cover how Angular decides that a component property value has changed,
```
kwalrath: Let's remove the promise and just say:

...: How Angular decides...

Is there anything we can point to? API docs or an example, maybe?

src/angular/guide/architecture.md
```diff
@@ -0,0 +1,592 @@
...
+when to update the screen, and how it uses **zones** to intercept asynchronous activity and run its change detection strategies.
+
+- **Events**: The events documentation will cover how to use components and services to raise events with mechanisms for
+publishing and subscribing to events.
```
kwalrath: This similarly could use a rewrite.

Is there anything we can point to? Maybe some API docs or an example?

src/angular/guide/architecture.md
chalin: Good. I've fixed the prose as you suggested except that I named the library just "angular2" (rather than "angular2/angular2.dart"). I also added links to the API pages.

src/angular/guide/architecture.md
chalin: Done

src/angular/guide/architecture.md
chalin: Neither we, nor the TS folks have any specific documentation on change detection. Come to think of it, I'd vote to simply drop this entry.

src/angular/guide/architecture.md
chalin: Again, how about we just drop it?

src/angular/guide/architecture.md
```diff
@@ -0,0 +1,592 @@
...
+- [**HTTP**](server-communication): Communicate with a server to get data, save data, and invoke server-side actions with an HTTP client.
+
+- [**Lifecycle hooks**](lifecycle-hooks): Tap into key moments in the lifetime of a component, from its creation to its destruction,
+by implementing the lifecycle hook interfaces.
+
+- [**Pipes**](pipes): Use pipes in your templates to improve the user experience by transforming values for display. Consider this `currency` pipe expression:
+
+      price | currency:'USD':true
+
+  It displays a price of 42.33 as `$42.33`.
```
chalin: @kwalrath - I'd be in favor of dropping the pipe example. It seems like excessive detail as compared to the level of detail that we give for every other entry in this list. WDYT? (I've commented it out for now).

chalin:
Updates done and staged copy has been refreshed too. Note that I commented out some entries from the "alphabetical list of other important Angular features and services".

src/angular/guide/architecture.md
kwalrath: I've just filed a bug about the imports: #787

src/angular/guide/architecture.md
kwalrath: Agreed, we should delete it.

PR #788, opened 2017-06-29
Resolve TravisCI build failure

mit-mit:


mit-mit:
Actually, nevermind, I thought Dart www was hitting this issue that I just fixed on Flutter:
https://github.com/flutter/website/pull/622

But doesn't look like it's hitting on Dart anyway; closing.

chalin:
Actually, we did hit that problem yesterday; I opened an issue over the firebase-tools: https://github.com/firebase/firebase-tools/issues/382

chalin:
Let's give your fix a try here too.

mit-mit:
kk, will wait and see if the two travis jobs pass before merging

chalin:
(The deployment script only kicks in when a commit is over the `master` branch, so I needed to merge to see if it actually works.)

chalin:
Yes, that worked. Thanks @mit-mit!

kwalrath:
Sounds like I should make this change to https://github.com/dart-lang/site-www/blob/master/deploy/deploy-firebase.sh.

chalin:
Yes, it will most likely be needed there too.

PR #795, opened 2017-07-05
Support page-specific social images

### LINKED ISSUE #37: Sharing image is ignored
If you try to share https://webdev.dartlang.org/codelabs/darrrt, it gets the default image (or on-page images), not the one specified in the header:

```
description: "Take your first step to learning Dart fast."
...
snippet_img: images/piratemap.jpg
```

(I think the image in the header is a broken URL, since it needs to be a full path, but still. It should link to https://webdev.dartlang.org/codelabs/darrrt/images/piratemap.jpg.)

kwalrath:
Fixes #37 

/cc @Sfshaza


src/_includes/head.html
```diff
@@ -24,7 +24,7 @@
   <meta property="og:description" content="{% if page.description | strip_html | strip_newlines | truncate: 160 %}{{ page.description }}{% else if page.excerpt %}{{ page.excerpt | strip_html | strip_newlines | truncate: 160 }}{% else %}{{ site.description | strip_html | strip_newlines | truncate: 160 }}{% endif %}" />
   <meta property="og:type" content="website" />
   <meta property="og:url" content="{{ site.url }}{{ page.url }}" />
-  <meta property="og:image" content="{% if page.image %}{{ page.image }}{% else %}{{ site.url }}{% asset_path 'dart-webdev-logo.png' %}{% endif %}" />
+  <meta property="og:image" content="{% if page.snippet_img %}https://webdev.dartlang.org{{ page.snippet_img }}{% else %}{{ site.url }}{% asset_path 'dart-webdev-logo.png' %}{% endif %}" />
```
filiph: Can we use `{{ site.url }}` instead of `https://webdev.dartlang.org`?  So, effectively, doing something like:

    <meta property="og:image" content="{{site.url}}{% if page.snippet_img %}{{ page.snippet_img }}{% else %}{% asset_path 'dart-webdev-logo.png' %}{% endif %}" />

src/_codelabs/darrrt/index.md
```diff
@@ -3,7 +3,7 @@ layout: codelab
 title: "Low-Level HTML Pirate Codelab"
 description: "Take your first step to learning Dart fast."
 permalink: /codelabs/darrrt
-snippet_img: images/piratemap.jpg
+snippet_img: /codelabs/darrrt/images/piratemap.jpg
```
chalin: We can avoid making these site-root paths. See my last comment below.

src/_includes/head.html
chalin: A few comments about this:
- Since this seems to be a local naming scheme, I'd vote for `og_image` as a page property name rather than `snippet_img` (which doesn't hint at what this page property is used for). Or possibly better: `og_image_path`.
- If the URL must be complete then I'd suggest using:<br>`{{site.url}}{{page.url}}/{{page.og_image}}`<br>instead of<br>`https://webdev.dartlang.org{{ page.snippet_img }}`.

chalin: One other point about the previous code: I don't think that it makes sense to use `asset_path` to uniquify the asset path since, I believe, we want the "og:image" URL to remain valid as long as possible -- in particular, not be invalidated every time we commit to site-webdev master.

To fix this we'd need to move the image in question to another folder that so that it is directly accessible.

kwalrath: Yes, site.url seems much better. And it'll work unchanged on www.

kwalrath: * Changing from `snippet_img` to `og_image` SGTM. (This code will be used on www.dartlang.org, as well, but only two pages were using it over there. And not successfully.)
* I'd rather specify complete paths (starting at `/`) to images, since some might be shared between pages in different directories. The less magic, the better. Starting at `/` seems fine because you don't want to depend on another site.
* `{{site.url}}` ftw.

kwalrath: Oh, and... removing asset_path SGTM. OK by you, @filiph?

chalin:
I researched `og:image` more carefully this morning and I have some suggestions, ... which I'll post shortly.  Can we hold off finalizing the PR until then?

src/_includes/head.html
filiph: Ok by me, thanks!

kwalrath:
#799 does all this and more.

PR #789, opened 2017-06-29
[WIP] Script and API link updates due to angular2 moving to `dart-lang/angular`

chalin:
Note that this PR **does not** rename the `angular2` package in the docs (since that is still the official package name under pub).

In this PR:
- Adjust build scripts so that they pickup the `dart-lang/angular` repo (since `angular2` is now just redirected to `angular`).
- The Angular API reference is now created from `dart-lang/angular`.
- Updated links to API entries to refer to `angular` instead of `angular2`.

Staged at https://angulardart-org-dev.firebaseapp.com

Contributes to #786
Fixes #226

PR #804, opened 2017-07-11
Script and API link updates due to angular2 moving to `dart-lang/angular`

chalin:
The `4-dev` branch counterpart to #789 (which was initially submitted over master).


chalin:
This PR wasn't auto-updating after a force-push, probably because the branch was already associated with #789. Closing, will use the new #805 instead.

PR #805, opened 2017-07-11
Script and API link updates due to angular2 moving to `dart-lang/angular`

chalin:
Supersedes #789 (over master), and #804 (over 4-dev).

Already reviewed via #789.

PR #806, opened 2017-07-11
Script and API link updates due to angular2 moving to `dart-lang/angular`

chalin:
Already reviewed via #789.
Supersedes #789 & #805 (over master), and #804 (over 4-dev).


kwalrath:
I'm worried about having `angular` (instead of `angular2`) visible in the API docs before a stable `angular` 3.x/4.x release is on pub. Where's the staged version of this change?

kwalrath:
Oh wait, I just noticed this is on the 4...dev branch. Never mind. :)

PR #807, opened 2017-07-11
Upgrade to stream_transform v0.0.6

natebosch:
Fixes https://github.com/dart-lang/stream_transform/issues/18

This version has a fix for `debounce` to forward events to all
listeners.

Also updates the test to start listening for the first value on the
stream before typing through the page object since typing also waits for
the Timer to fire.

### DIFF
--- a/examples/ng/doc/toh-6/pubspec.yaml
+++ b/examples/ng/doc/toh-6/pubspec.yaml
@@ -9,7 +9,7 @@ environment:
 dependencies:
   angular2: ^3.1.0
   http: ^0.11.0
-  stream_transform: ^0.0.5
+  stream_transform: ^0.0.6
   # #enddocregion additions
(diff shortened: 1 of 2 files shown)


chalin:
Thanks for the PR! LGTM, except that one of the Travis jobs is failing.

I don't mind fixing it. If you'd like to give it a try, then run these two commands from the repo folder:
- `gulp create-example-fragments --fast`
- `./scripts/check-code-excerpts.sh`

And then submit the updated file(s). If you'd rather I do this tweak, let me know.

chalin:
Closed by #808. As explained in #808, I don't have the necessary permissions to edit your PR/branch so I submitted another PR. Thanks again Nate!

natebosch:
Ah cool - thanks for getting it in :)

Issue #621, opened 2017-05-13
[ng] write component testing page


chalin:

chalin:
A first iteration of the component testing docs have been published.

PR #810, opened 2017-07-13
[ng] toh test updates: method rename and added docregions

chalin:
This PR does not impact the prose.

Contributes to #621


chalin:
Travis is green, this has no user-visible impact on the site docs, and the changes are not significant, so I'll go ahead and merge.

PR #821, opened 2017-07-15
chore: ACX version upgrade; delete tmp file

chalin:


chalin:
Note that this version upgrade only affects the version used to generate the ACX API docs. No changes have been made to example code (which will pickup the new version w/o any pubspec changes). Merging.

Issue #831, opened 2017-07-18
[4-dev] Update versions page


chalin:
This is a followup to https://github.com/dart-lang/site-webdev/pull/827#issuecomment-316139954.

We need to decide how to populate the [versions table for the 4-dev](https://webdev-dartlang-org-dev.firebaseapp.com/angular/version) branch. 

On master, the versions table reads:
> ![screen shot 2017-07-18 at 1 47 44 pm](https://user-images.githubusercontent.com/4140793/28331626-caccd812-6bbf-11e7-80c8-0da9a4545c86.png)

We can easily change the site config variables so that 4.0 appears at the top of pages by assigning 4.0 to the "stable version" variable(s), but that will cause the "Stable" row entry of the versions table to contain 4.0.0-alpha (and the "Dev" row to be empty). Is that ok with you, @kwalrath?

kwalrath:
Hmmmm... Would it be possible to have text that only shows up when you aren't on webdev.dartlang.org/master? Then we could put a big warning in this page that it's a pre-release version of the docs, and look at webdev.dartlang.org/angular/version for the latest info about versions.

Heck, I think I want this warning on all the pages, somehow...

chalin:
Something I've been wanting to suggest for a while now, is to include the ng version number in the main nav and the sidenav. We could name the version "4.0 PREVIEW". That would make ensure that the version is visible on all pages. WDYT?

kwalrath:
SGTM!

chalin:
Done via rebase of master (after #835 was merged).

PR #835, opened 2017-07-18
[ng] Add NG major version to main nav, update Version page, etc

### LINKED ISSUE #834: Add AngularDart version to main nav and sidenav
Followup to https://github.com/dart-lang/site-webdev/issues/831#issuecomment-316148683.

chalin:
Fixes #834. Contributes to #831.

- Adds the major version number to AngularDart in the main nav.
- Do NOT add any version numbers to the sidenav (it is complicated to do so, and probably not worth it at this point).
- If the current version of angular is an alpha or beta release, then:
  - A warning will appear in the site page banner (because the main and side navs do not appear on narrow displays, and also because it is much more visible).
  - On the version page, the current release will be qualified as "Current" rather than "Stable".
- Version page:
  - The "Dev" row in the version table only appears if there is a dev version of either NG or ACX.
  - The "Alpha is production quality" comment appears if any version displayed in the versions table contains "alpha".

Staged at https://angulardart-org-dev.firebaseapp.com/angular


src/_includes/navigation-main.html
```diff
@@ -5,7 +5,7 @@
   </a>
   <ul>
     <li><a href="/guides/get-started">Get Started</a></li>
-    <li><a href="/angular">AngularDart</a></li>
+    <li><a href="/angular">AngularDart {{site.ng.vers.major}}</a></li>
```
kwalrath: I'm not sure about having the version # in the main nav. It reminds me of the big switch from Angular to Angular 2, which is probably not the message we want to give. I'd be more comfortable with `site.ng.vers.abbr`, but actually... maybe we only want the version called out in the /angular & /angular_component headers (as they are now), in the version page, and in the warning on the dev version of the site.

chalin: Ok. I'll put that out from the main nav.

src/angular/version.md
```diff
@@ -17,63 +17,78 @@ packages. These packages often have a more recent development _(dev)_ release.
...
+        </a>
+      </div>
       {% endif %}
     </td>
   </tr>
+  {% endif %}
 </table>
 
-<!-- Comment out when we're not in alpha -->
+{% if site.ng.vers.full contains "alpha" or site.ng.dev.version contains "alpha" or site.acx.dev.version contains "alpha" %}
```
kwalrath: Why is the "Angular alpha releases are production quality" heading appearing in the staged version? (https://angulardart-org-dev.firebaseapp.com/angular/version)

I actually don't think it's a bad thing to show everywhere, but I didn't expect it to show up.

src/angular/version.md
```diff
@@ -17,63 +17,78 @@ packages. These packages often have a more recent development _(dev)_ release.
...
+      <div>
+        <a href="{{site.dev.url}}/angular/guide">
+          {{site.dev.url | regex_replace: '^https?://' }}<b>/angular/guide</b>
+        </a>
+      </div>
+      {% endif %}
+      {% if site.acx.dev.version %}
+      <div>
+        <a href="{{site.dev.url}}/components">
+          {{site.dev.url | regex_replace: '^https?://' }}<b>/components</b>
```
kwalrath: We should probably add the class that makes the external link indicator go away.

src/angular/version.md
```diff
@@ -17,63 +17,78 @@ packages. These packages often have a more recent development _(dev)_ release.
...
+      <div>
+        <a href="https://pub.dartlang.org/packages/angular/versions/{{site.ng.dev.version}}#pub-pkg-tab-changelog" class="no-automatic-external">
+          angular <b>{{site.ng.dev.version}}</b>
+        </a>
+      </div>
+      {% endif %}
+      {% if site.acx.dev.version %}
+      <div>
+        <a href="https://pub.dartlang.org/packages/angular_components/versions/{{site.acx.dev.version}}#pub-pkg-tab-changelog" class="no-automatic-external">
+          angular_components <b>{{site.acx.dev.version}}</b>
```
kwalrath: Not new, I think, but in many screen sizes, the "angular_components" entry goes to 2 lines, which makes it not align with the documentation links.

src/angular/version.md
kwalrath: `class="no-automatic-external"`

src/angular/version.md
kwalrath: Oh, I get it. If there's any alpha visible at all this appears. I withdraw my question!

chalin: Because the ng.dev.version contains "alpha". If none of the displayed versions contain "alpha" then that paragraph is dropped.

src/angular/version.md
chalin: Yeah, I'd seen that but figured it was ok (for now). I didn't want to have to start playing with multispan rows (although I could use an nbsp between the package and version).

chalin: Let me try nbsp.

kwalrath: We could also just file a beginner, help wanted bug...

PR #851, opened 2017-07-20
[ng] API docs: include top level angular2/angular2 package

chalin:
This is a precursor to eventually dropping angular2.common.


chalin:
Actually, let's only do this for the 4-dev branch.

Issue #852, opened 2017-07-20
Fix issues in angular/api homepage


kwalrath:
I noticed a few issues in https://webdev.dartlang.org/angular/api (in addition to #787):

* [ ] Deprecated APIs aren't crossed out. For example, EventEmitter and View should be crossed out.
* [x] Some rows have nothing in the 1st & 2nd columns (just the 3rd). For example, under angular2.core:
![image](https://user-images.githubusercontent.com/2164483/28437383-41886e04-6d4f-11e7-888b-be9d3acd4a9e.png) Fixed by #853
* [x] Some API items are uncategorized. For example, in the above screenshot, GetterFn & MethodFn have no icon. A more complete list: BaseHRefFromDOMProvider, baseHRefFromDOM, TrackByFn, reflector, SetterFn, MethodFn, GetterFn, ComponentStateCallback, ValidatorFn, TouchFunction, ChangeFunction. Fixed by #853

chalin:
All bugs have been fixed. I've moved the remaining item to #929.

Issue #670, opened 2017-05-31
[ng] 4.0 prep


chalin:
- [x] [4.0.0-alpha+3 changes](https://github.com/dart-lang/angular/blob/master/angular/CHANGELOG.md#400-alpha3)
  - [x] `COMMON_DIRECTIVES` &rarr; `CORE_DIRECTIVES, formDirectives`
  - [x] Add `angular_forms/angular_forms.dart` to files using `formDirectives`
- [x] [4.0.0-alpha+2 changes](https://github.com/dart-lang/angular/blob/master/angular/CHANGELOG.md#400-alpha2):
  - [x] `package:angular/router.dart` &rarr; `package:angular_router/angular_router.dart`
  - [x] `angular/platform/common.dart` and `angular/platform/common_dom.dart` &rarr; `angular/angular.dart`
  - [x] Rename `FORM_DIRECTIVES` to `formDirectives`.
  - [x] Fix ddc compiler errors in examples #703, re-enable skipped tests.
  - [x] Replace injected `ElementRef` by `Element` #948
- [x] [4.0.0-alpha+1 changes](https://github.com/dart-lang/angular/blob/master/angular/CHANGELOG.md#400-alpha1)
- [x] [4.0.0-alpha changes](https://github.com/dart-lang/angular/blob/master/angular/CHANGELOG.md#400-alpha):
  - [x] replace `/deep/` (and `>>>`) by `::ng-deep` #949
  - [x] Remove references to `resolved_identifiers` #824
  - [x] Adjust component-styles example now that [CSS @import is working](https://github.com/dart-lang/angular2/commit/1e276958988606d67100ff1e5356cd10a8976d6a) - #669
  - [x] Enums and globals can now be accessed in templates. Fix docs and examples (e.g. template-syntax). Cf. https://github.com/dart-lang/angular2/issues/374
   - [x] address package rename/repo relocation for angular2 and angular_test #786

Other:
- [x] #880: add API docs to sitemap?
- [x] fix sitemap API entries - followup to https://github.com/dart-lang/site-webdev/pull/888#issuecomment-319745549
- [x] update redirect for angular2.common*
- [x] ensure `github_issues` app is running on latest NG version, https://github.com/dart-league/ng_bootstrap/issues/74, https://github.com/dart-league/ng_bootstrap/issues/78
- [x] from https://github.com/dart-lang/site-webdev/pull/853#issuecomment-316862196: 
  > request an immediate re-indexing when we do change the [API] URLs

[...]

PR #853, opened 2017-07-20
[ng] API homepage fixes; added angular2 pkg, added new filter categories, etc

chalin:
Solves all issues of #852 except for marking items as deprecated.

Staged at https://angulardart-org-dev.firebaseapp.com/angular/api


kwalrath:
Looks great! We should add `[T] Typedef` to the "type" list, right?

chalin:
Oh, I thought that was there! Thanks for catching that.

tools/api-builder/dart-package/services/preprocessDartDocData.js
```diff
@@ -41,7 +41,8 @@ module.exports = function preprocessDartDocData(log, dartPkgConfigInfo) {
           const matches = e.href.match(/-([a-z]+)\.html/);
           let type = matches ? (e.typeFromHref = matches[1]) : e.type;
           // Conform to TS type names for now.
```
kwalrath: Delete this comment?

chalin:
> How hard would it be to implement redirects for things that used to be in angular2.common?

Depends at what granularity you'd like for it to be handled. Will you want redirects once we switch to 4.x too?  If so, do you want to open an issue for that?

chalin:
Last commit just removed an obsolete comment. Travis was green before that. Merging.

kwalrath:
>> How hard would it be to implement redirects for things that used to be in angular2.common?
>
> Depends at what granularity you'd like for it to be handled. Will you want redirects once we switch to 4.x too?  If so, do you want to open an issue for that?

Let me look into this a bit. I'm not sure it's worth the trouble.

tools/api-builder/dart-package/services/preprocessDartDocData.js
chalin: Yes, thanks. Done.

kwalrath:
Looking at the analytics, the vast majority of API doc visits come in from Google searches, so once the Google index updates redirects would be irrelevant.

It'd be good to request an immediate re-indexing when we do change the URLs.

chalin:
FYI, I've added a reminder to #670 concerning the re-indexing.

Issue #787, opened 2017-06-28
In Angular API reference, be clear about what import to use.


kwalrath:
FIXED this part:
> For example, when you go to the AngularDart API homepage (https://webdev.dartlang.org/angular/api), it doesn't say what you should import to get the APIs listed under angular2.common._

This info should be—at the very least— on both on https://webdev.dartlang.org/angular/api and the individual library page (in this case, https://webdev.dartlang.org/angular/api/angular2.compiler/angular2.compiler-library). You can't get to the library page from the API homepage, but you can get there by clicking the library name in the header.

Strangely, once you do that, "angular2.common" is crossed out in the header and left nav, which makes it look like it's deprecated. How can the first library in the API page be deprecated?!

chalin:
FYI, [`angular.common` is gone as of `4.0.0-alpha+1`](https://github.com/dart-lang/angular/commit/6e55e2666207f96634290ffbb97b7932be3a3597#diff-1f554f358a75ba206c49542dd0c50aabR45).

chalin:
There have been considerable improvements made to dartdoc recently (thanks @jcollins-g!). In particular, canonical reference resolution. Because of these improvements, and because (as of 4.x) our API page will be showing `angular` package entries (which we used to hide to avoid duplicates), I believe that the issue reported here has essentially been resolved.

For a preview of the 4.x API page see https://webdev-dartlang-org-dev.firebaseapp.com/angular/api.

I believe that this issue has been resolved in that, for any given symbol X appearing under the package grouping P, one would need to `import 'package:angular/P.dart`.

Given this general rule, we could add a note to this effect on the API page, or put the import statement as a hover/tool-tip anchored on the package name, ... .  

Kathy, WDYT?

kwalrath:
I'd like to have the import statements explicitly in this page (and elsewhere, if possible), since hover/tooltips aren't good for mobile, and the clearer we can make this doc, the better. Let's keep this issue open until we have visible help on importing.

I'll file issues for the other problems I've noticed in our API docs.

chalin:
Here is a proposal for showing imports on the API homepage (I'm only showing Functions so that you can see multiple groups -- note that the group heading can be a link to the library dartdoc API page):
> ![screen shot 2017-07-21 at 5 53 10 pm](https://user-images.githubusercontent.com/4140793/28483899-8c56592a-6e3d-11e7-9ff5-b43d481a1fe0.png)

[...]

PR #861, opened 2017-07-21
[ng] API homepage: show import statement for each library

chalin:
Contributes to #787 -- see https://github.com/dart-lang/site-webdev/issues/787#issuecomment-317121588 for screenshots.

Staged at https://angulardart-org-dev.firebaseapp.com/angular/api (you might have to clear the cache and/or open in an incognito window).


src/resources/js/directives/api-list.js
```diff
@@ -46,7 +46,15 @@ angularIO.directive('apiList', function () {
       '      ' +
       '<article class="l-content-small docs-content">' +
       '  <div ng-repeat="section in $ctrl.groupedSections" ng-if="$ctrl.filterSections(section)" ng-cloak="ng-cloak">' +
-      '    <h2>{{ section.title }}</h2>' +
+      '    <a ng-href="/angular/api/{{section.title}}/{{section.title}}-library" target="_blank" rel="noopener">' +
+      '      <h2>' +
+      '        {{section.title}}' +
+      '          <span style="vertical-align:middle;">&#8227</span>' +
```
kwalrath: I love it except for the icon, which looks like it should open something. Other possibilities:

* nothing (except maybe more space)
* a colon (:)
* a non-triangular arrow such as →

kwalrath: The mobile version shouldn't have an icon. And I think we really don't need one in the desktop version either. If that looks to weird, we could put it below the library name, as in the mobile version.

chalin: Ok. let's try dropping the triangle in favor of a little more space.

chalin:
Changes made:
- There is now only space between the library name and its import statement.
- Only the library name is a link to its API entry
  The import statement, no longer being a link, is more easily selectable.

Staged server refreshed: https://angulardart-org-dev.firebaseapp.com/angular/apiStaged at https://angulardart-org-dev.firebaseapp.com/angular/api

Here's the updated look:
> screen shot 2017-07-22 at 7 18 38 am

> screen shot 2017-07-22 at 7 19 24 am

Issue #880, opened 2017-07-28
Prepare for changed API URLs when 4.0 is released


kwalrath:
When 4.0 is released, angular2 → angular, which will cause all the URLs under angular/api to change. For example:

* `https://webdev.dartlang.org/angular/api/angular2/NgFor-class` → `https://webdev.dartlang.org/angular/api/angular/NgFor-class`
* `https://webdev.dartlang.org/angular/api/angular2.core/OnChanges/ngOnChanges` → `https://webdev.dartlang.org/angular/api/angular.core/OnChanges/ngOnChanges`

We'll want to update the Google search index ASAP and possibly redirect appropriately.

We'll also need to redirect from angular2.common* to angular (followup to #888).

This is similar to but larger than the angular2.common -> angular2 change, which caused a fair number of 404s. See #877 for more information.

kwalrath:
I think our best bet is to add the API pages to the sitemap. @chalin could you help me with that?

kwalrath:
Related tasks are in #670.

Issue #877, opened 2017-07-28
Site search still generates URLs to angular/api/angular2.common which are 404s


michschn:
URL: https://webdev.dartlang.org/angular/api/angular2.common/NgClass-class
Expected page: https://webdev.dartlang.org/angular/api/angular2/NgClass-class

The bad URL was generated from the site search

see 1st result:
https://webdev.dartlang.org/search?cx=011220921317074318178%3Ata2gkewbtuu&ie=UTF-8&hl=en&q=ngClass

kwalrath:
Thanks for the report. We didn't do redirects because most of those pages are found via the index or Google search; I'm going to try again to get Google to update its search index.

kwalrath:
7/24 I requested a recrawl of /angular/api and /angular/api/angular2/NgFor-class. I thought that might cover everything, but although search results for NgFor are correct, search results for NgClass still aren't.

I've requested a recrawl of NgClass. I'll go through the analytics to find other pages people visit most often, and recrawl them.

Keeping this bug open until I confirm that NgClass, at least, is indexed correctly.

kwalrath:
Firebase doesn't seem to have very expressive redirect support, but I could do an interim fix by adding something like this to `firebase.json`:

```
     { "source": "/angular/api/angular2.common/**",
       "destination": "/angular/api/angular2/angular2-library",
       "type": 302
     },
```

kwalrath:
I wonder if adding the API pages to https://webdev.dartlang.org/sitemap.xml might help.

kwalrath:
As an experiment, I went to the Search Console > Google Index > Remove URLs and asked to temporarily hide https://webdev.dartlang.org/angular/api/angular2.common. Status is Pending.

chalin:
@kwalrath - let me know how that works out. (If needed, we could just add a redirect rules for each (top-level) entry that used to be under https://webdev.dartlang.org/angular/api/angular2.common; there aren't that many. Of course, we might want to have a better solution when we migrate to 4.x).

kwalrath:
I'm now seeing a mix of angular2.common and angular2 results.

kwalrath:
I just checked the Search Console. The status of https://webdev.dartlang.org/angular/api/angular2.common is "Removed", but that didn't really seem to help search results. I just requested that /angular/api/angular2/angular2-library and its directly linked URLs be re-indexed (I'd wanted to do that last time, but I ran out of requests).

[...]

PR #887, opened 2017-08-02
chore: add NG API top-level library entries to sitemap

chalin:
Contributes (fixes?) #877, #880.


chalin:
Here's the issue for adding the components/api entries: #889.

PR #911, opened 2017-08-08
Fix wrapping of Angular page titles

### LINKED ISSUE #345: Fix wrapping of Angular page titles
From a review by @chalin of #344 (screenshots are from https://webdev.dartlang.org/angular/guide/):

> ...The old CSS resulted in "clean" title wrapping for narrow displays, like this:
>> screen shot 2017-01-28 at 10 58 00 am
>
> On the other hand, the new titles look like this:
>> screen shot 2017-01-28 at 10 57 44 am

Patrice suggested going back to a background image, but I don't know how to do that while making the image a link. I tried a bit to make this work (see the following code) but it didn't, for some reason:

```
.ng-page-title {
  margin-left: 3.1em;
  text-indent: -5.3em;
}
```

kwalrath:
Fixes #345 

Staged at URLs like https://kw-webdev-dartlang-1.firebaseapp.com/angular/guide. (Make your window very narrow; compare to https://webdev.dartlang.org/angular/guide.)

Uses a CSS+HTML solution suggested by @cgomezmendez (thanks!). I'm not sure whether my particular implementation is the best, though.


src/_layouts/angular.html
```diff
@@ -16,10 +16,10 @@
           {% include breadcrumbs.html %}
           <div class="ng-page-title">
             <h1>
-              <a href="/angular"><img src="/angular/images/angulardart.svg"></a>
+              <a href="/angular"><img src="/angular/images/angulardart.svg" style="float:left"></a><p>
```
chalin: This isn't valid HTML5: a [heading element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Heading_Elements) cannot contain a paragraph:
```
Error: Element p not allowed as child of element h1 in this context. 
```

chalin: Try:
- Putting the image and anchor outside of the `<h1>`
- Dropping the paragraph element

kwalrath:
I tried moving the image+anchor outside the h1, but the wrapping went bad again.

@cgomezmendez do you have any alternative suggestions?

chalin:
Change the `<p>` to a `<span>` works:
> screen shot 2017-08-09 at 07 10 07

chalin:
Actually, that doesn't fix the wrapping for the version number:
screen shot 2017-08-09 at 08 34 40

There are too many elements for the title. Here is something that fixes the wrapping for both the title text and the version; drop the `<div>` and fall back to only using `<h1>` like this:
```html
<h1 class="ng-page-title">
  <a href="/angular"><img src="{% asset_path angulardart.svg %}"></a>{%
  assign title = page.long-title | default: page.title | append: ' ' | append: page.status
  %}{{ title }}
  <a href="/angular/version">{{site.ng.vers.abbr}}</a>
</h1>
```
```css
.ng-page-title {
  img {
    float:left;
    margin-right: $unit * .5;
    width: $unit * 5;
  }
  a {
    font-size: $font-size-base;
  }
}
```
Which results in proper indentation and wrapping; e.g.:
> screen shot 2017-08-09 at 08 38 21

> screen shot 2017-08-09 at 08 38 29

A few remarks, and notes about other changes:
- We need to use `assert_path` for the image, as shown above.
- `angulardart.svg` will need to be moved to `src/_assets/images`.
- Please fix the ACX layout to match the changes to the page title made to the NG layout.
- Also, the ACX layout version variable needs to be changed (my bad for not having updated it when I renamed those variables).

kwalrath:
@chalin PTAL

kwalrath:
I'm tempted too, but maybe for another day...

PR #917, opened 2017-08-08
Weird hack to make pub links work again

kwalrath:
Go to https://kw-webdev-dartlang-2.firebaseapp.com/angular/version. Click the link to 	
**angular 4.0.0-alpha** (in the Dev row of the table). It should take you to a page under pub.dartlang.org.

If you do the same in https://webdev.dartlang.org/angular/version, you stay on webdev.dartlang.org unless you tell the browser to open the link in a new window. (This is on Chrome. I haven't tested other browsers.)

/cc @kevmoo


kevmoo:
QQ: do we have any crazy JS on these pages that's catching link clicks or anchors or anything?

kevmoo:
@chalin @kwalrath I did a bit of debugging through the main.js file – I think it's catching the clicks here and doing weird things.

Couldn't figure it out through the minification – but it'd be something to look at

chalin:
> ... do we have any crazy JS on these pages

Yes, there is both jQuery and AngularJS happening on the Angular pages.

kwalrath:
Shipping it now, but I'll file a bug about fixing this for real.

PR #922, opened 2017-08-09
Generate package versions table from data file

### LINKED ISSUE #920: Refactor versions table and generate it from package data
As of 4.x, certain Angular libraries will be distributed as separate packages, each with their own version numbers. The versions table needs to be refactored to accommodate this, and the table generation should be data driven.

Subsumes #836.

chalin:
Fixes #920 

Staged at https://angulardart-org-dev.firebaseapp.com/angular/version


kwalrath:
Looks great! Some nits:

* The link to 4.0.0-alpha+2 doesn't work because of the angular2 -> angular name change.
* Shouldn't the angular_test row have a link to a stable version/docs? Is the reason there's no doc link in the Dev version because you haven't brought those changes into your dev branch yet?
* Could we use CSS to have ellipsis at the beginning/middle of table entries (e.g. angular_components) when they're too wide for their cell?

chalin:
> - The link to 4.0.0-alpha+2 doesn't work

I've provided a temporary fix. As of 4.x we'll be able to drop this little "hack".

> Shouldn't the angular_test row have a link to a stable version

The first release of `angular_test` is 1.0.0-alpha.
We don't host docs for `angular_test` (at least not yet), that is why there is no docs link.

> Could we use CSS to have ellipsis at the beginning/middle of table entries

Yes we could, but ellipsis can be tricky to get working. Would you mind if we did that in a follow-up PR?

kwalrath:
I thought that angular_test might link to https://webdev.dartlang.org/angular/guide/testing/component. The other doc links go to non-API pages, and I figure this page is a good one for showing some context for angular_test.

(I just realized that that page has at least 3 links to the angular_test pub page. I'll try to reduce the number when I edit that doc...)

chalin:
I've added ellipsis; e.g.
>  screen shot 2017-08-09 at 16 21 10

chalin:
Docs link added for `angular_test`.

kwalrath:
Looks great!

PR #951, opened 2017-08-18
changelog updates for `ElementRef`, `/deep/`, `export`

chalin:
Contributes to #670


src/angular/changelog.md
```diff
@@ -84,6 +84,9 @@ Ensure that the Liquid variables below are properly hard-coded once 4-dev become
   * `COMMON_DIRECTIVES` &rarr; `CORE_DIRECTIVES, formDirectives`
   * `const Provider(x,y)` &rarr; `const Provider<T>(x,y)` for a provider of `T` instances;
      this is a first step towards [strongly-typed providers](https://github.com/dart-lang/angular/issues/407)
+  * Replaced injected `ElementRef` by `Element`, which requires an import of `dart:html`
```
kwalrath: Instead of "Replaced...by" I'd say "Changed...to"

src/angular/changelog.md
```diff
@@ -84,6 +84,9 @@ Ensure that the Liquid variables below are properly hard-coded once 4-dev become
...
+  * Replaced the CSS pseudo selector `/deep/` by `::ng-deep`
```
kwalrath: Replaced...by -> Changed...to

src/angular/changelog.md
```diff
@@ -84,6 +84,9 @@ Ensure that the Liquid variables below are properly hard-coded once 4-dev become
...
+  * Used `@Component` `export` for enums
```
kwalrath: I don't know what this means. Could you give a little more information or an example?

chalin: How about if I add a reference to https://github.com/dart-lang/angular/issues/374? The change was to a `template-syntax` component. The export was used to make an enum available in the component template.

kwalrath: That's a lot to read through. Maybe just a link to the diff in our code, plus wherever it's covered in the angular changelog? (Actually, I can't find it in https://pub.dartlang.org/packages/angular/versions/4.0.0-alpha%2B3#pub-pkg-tab-changelog. Should we file a bug to add it, or did I just miss it?)

chalin: It is the first item under "New features" of 4.0.0-alpha.

kwalrath: Ah, I was looking for "enum". So maybe what we want to say is something like:

Changed components to use the new `exports` parameter of `@Component` to publish enums, instead of using an alias getter in the component class.

(If indeed that's what you did. I don't know where this change occurred, which is why I'm floundering a bit.)

src/angular/changelog.md
chalin: done

src/angular/changelog.md
chalin: done

src/angular/changelog.md
chalin: +1. I've tweaked your suggestion a bit further, and also added a link to the change as well as to the RFC.

PR #955, opened 2017-08-19
Create a single unified API reference

chalin:
- Introduces a single unified API reference at /api.
- It replaces /angular/api and /components/api, which have been dropped.
- Redirects old API URLs to /api.
- Sidenav now has a single "API Reference" entry under Resources.
- Also added Changelog to the sidenav, just below that.

This **same infrastructure** will be used to create a unified API reference under the **4-dev branch** where Angular has been partitioned into angular (proper), angular_forms and angular_router.

Staged at https://angulardart-org-dev.firebaseapp.com/api

A screenshot is given below. Note that the import statements are still being generated.

----
screen shot 2017-08-19 at 15 17 38


chalin:
After trying out a few approaches, I actually prefer this solution. URLs are of the form: **/api/\<package>/library**, which allows us to preserve the package API doc landing pages. E.g.,
- https://angulardart-org-dev.firebaseapp.com/api/angular2
- https://angulardart-org-dev.firebaseapp.com/api/angular_components

kwalrath:
Just curious, since this is at the top level now... Would it be possible to add dart:html (and other web-specific APIs) to this page (similar to how https://docs.flutter.io/ has dart:async, dart:core, etc., as well as the Flutter APIs).

Which makes me wonder, where would we stop? In theory, we could show everything but dart:io and dart:mirrors, but I'm not sure how helpful that would be. For AngularDart programmers, dart:core, dart:convert, and dart:async might be the most helpful libraries to feature.

Would it be easy to play with this? (If so, I could mock up a few solutions.)

The other alternative is to keep the same libraries on this page but add links to other relevant API docs.

chalin:
> Would it be possible to add dart:html

Yes, though the mechanism for creating and merge it will be different from pub packages. (If it were another pub package then we could just add it to the list.)

> If so, I could mock up a few solutions.

Yes, I can do that first thing tomorrow.

Is this PR good to merge?

kwalrath:
I think we need a bit of intro text in the api page before we merge this. Just something to say what APIs are on this page and where they can find other APIs they can use in their web development.

chalin:
The [staging server has been updated](https://angulardart-org-dev.firebaseapp.com/api); it now includes the following SDK libraries: async, core, convert, html.

Note that for the SDK libraries:
- Entries link to `api.dartlang.org` (currently the latest stable release).
- I've removed the import statements for the SDK libraries.

Here is what it might look like:
screen shot 2017-08-22 at 12 04 23

kwalrath:
The version you just posted looks great! The filters overlap a bit when the screen is small, but are still usable, so I think we could ship this and improve the layout later.

Before we merge this, we'll need some basic redirects, right?

chalin:
> The filters overlap a bit when the screen is small

I couldn't find a quick fix. This will require further investigation.

> we'll need some basic redirects, right?

Right. They are already in. You can give them a try from the staging server.

kwalrath:
> > we'll need some basic redirects, right?
> 
> Right. They are already in. You can give them a try from the staging server.

Sorry, missed that! Funnily, the first URL I tried was for NgModel (our top API landing page, aside from angular/api), which made me realize that its docs appear in both angular2 & angular_components. But as far as I can tell, that's the only API that's duplicated.

We could get a bit fancier and filter according to the package they were looking for:

/angular/api/** -> /api#!?package=angular2
/components/api/** -> /api#!?package=angular_components

I'm not sure that always works, though.

LGTM!

chalin:
Here is the fixed wrapping:
> screen shot 2017-08-22 at 19 47 32

> screen shot 2017-08-22 at 19 48 01

> screen shot 2017-08-22 at 19 48 23

chalin:
> /angular/api/** -> /api#!?package=angular2
> /components/api/** -> /api#!?package=angular_components

Done.

PR #973, opened 2017-08-25
Add "API home" button to API entry pages

### LINKED ISSUE #959: Improve API reference navigation
There's no easy way to get from the new API docs to the API homepage (or anywhere else on webdev.dartlang.org).

We could work around this using text in the package READMEs and library doc comments, but it'd be better to have this automated, somehow.

chalin:
Fixes #959

Staged at: https://angulardart-org-dev.firebaseapp.com/api


chalin:
> Looks great! Let's get it out there, and we can tweak it as needed.

👍 

> Are you going to clean out the commented out lines of code?

I'll leave them in for now (as a record of an alternative approach), which I'll clean out in a subsequent PR.  I chose the current solution because it gives a slightly better positioning of the icon.

Issue #975, opened 2017-08-25
Add another sidenav section


kwalrath:
The RESOURCES part of our current sidenav (pictured below) has a couple of problems:

* It contains 2 angular-specific resources. The other resources are more general web development resources. The angular-specific resources:
  * [API Reference](https://webdev.dartlang.org/api): has an angular icon although it's top-level; contains only angular* classes
  * [Changelog](https://webdev.dartlang.org/angular/changelog): has an angular icon, is under the /angular path; it theoretically could contain non-angular changes... perhaps we should move it.
* It's too long to be easily scanned.

Current sidenav: 

![image](https://user-images.githubusercontent.com/2164483/29734955-2a650e9a-89aa-11e7-8dd7-3243a540eb2b.png)

Possible new sidenav (removes links to Articles, Glossary, & one Codelab):

ANGULAR DART
About
Guide >
Tutorial > 
Advanced >

ANGULAR DART
COMPONENTS
About
Example

ANGULAR DART
REFERENCE
API Documentation
Changelog
Cheat Sheet

RESOURCES
Codelabs
Community and Support
Libraries
Tools
Tutorials

chalin:
Good suggestion. Here is a variant staged at https://angulardart-org-dev.firebaseapp.com as it would look like on a phone:
> <img width="414" alt="screen shot 2017-08-26 at 09 15 37" src="https://user-images.githubusercontent.com/4140793/29741646-29ef08e0-8a3f-11e7-81cd-005e18c03925.png">

I think that it is important to keep a direct link to the Glossary.

We can further reclaim vertical real estate by:
- Dropping the "About" entries and making the headers link. (Yes, we've discussed this before :)
- By adopting the previous suggestion, we could then make "Get Started" a linked top-level entry (as opposed to an overly padded button :)

kwalrath:
I like your ideas for reclaiming vertical real estate, and with those changes I'm fine with keeping Glossary in the sidenav. (I just looked at the analytics, and it's the 61st most visited page on webdev, behind 13 of 17 always-visible sidenav pages; changelog, articles, & community have fewer views. I suspect the glossary wouldn't get nearly as many visits if it weren't in the sidenav.)

Do you have any objection to shortening and alphabetizing the RESOURCES links, as in my original suggestion above?
 
Finally, we should probably change site-www to reflect whatever we do here.

chalin:
> I like your ideas for reclaiming vertical real estate

[...]

PR #977, opened 2017-08-26
Sidenav adjustments, including adding "Reference" section

chalin:
Contributes to #975

Staged at https://angulardart-org-dev.firebaseapp.com


src/_includes/linklist.html
```diff
@@ -1,24 +1,36 @@
-{% for link in linklist  %}
-  <h4 {% if page.url contains link.permalink %}class="active"{% endif %}>
-    {{ link.title }}
+{%- comment %}Strip out any trailing '/', etc. for exact URL match.{% endcomment -%}
```
kwalrath: I'm curious, why the added `-`s?

chalin: It strips the whitespace from before/after the Liquid directive.

kwalrath:
Looks great, except for the differences between angular CSS and regular site CSS. E.g. compare the sidenavs of https://angulardart-org-dev.firebaseapp.com/tools and https://angulardart-org-dev.firebaseapp.com/api.

chalin:
Yes, but a lot more work would be needed to harmonize the main site styles with the Angular specific styles. Possibly something for another day.

kwalrath:
I've filed #980 about the style differences.

PR #993, opened 2017-08-31
Master is now for 4.x docs, 5-dev for the next dev release, fix changelog

chalin:


chalin:
Going to merge now so that we can get the new 4.0 site auto-deployed.

PR #996, opened 2017-09-01
Refresh code excerpts and diffs. Adjust refresh check to include all src

chalin:


src/playground/index.md
```diff
@@ -16,20 +16,29 @@ Quickstart pubspec w/ title:
...
-    var [!name!] = 'Angular';
+    var name = 'Angular';
   }
 ```
 
 <?code-excerpt path-base="quickstart"?>
 
 Quickstart pubspec w/ title:
-<?code-excerpt "pubspec.yaml" title?>
+<?code-excerpt "lib/app_component.dart" title?>
```
kwalrath: Probably doesn't matter since this is just the playground, but the description and the code don't match.

chalin: I've cleared out the playground page (#1017).

PR #997, opened 2017-09-01
Add "Previous" release column on version page

chalin:
I suspect that folks might want to compare docs with the previous release. This adds links to the 3.x docs. It would look like this:

> screen shot 2017-08-31 at 21 18 01


kwalrath:
Good idea! (Sorry for the delay; I thought I'd already reviewed this.)

chalin:
np

PR #1014, opened 2017-09-03
gulp/dartdoc task: by default, don't regenerate docs

chalin:
Now that we aren't switching back and forth between 3.x and 4.x docs, by default don't regenerate dartdocs (we used to have to include `--fast` to achieve that behavior).


kwalrath:
@chalin should we update the README?

chalin:
PTAL at #1035

PR #1036, opened 2017-09-07
chore: properly regenerate API docs on a `build --clean`

chalin:
There was an issue with the order of initialization of `config.dartdocProj` when `--clean` is used locally. (Followup to #1014)


kwalrath:
Oh, is this issue why your recently staged PR didn't have API docs?

chalin:
Possibly, though there have been times where I haven't bothered to rebuild the API docs before staging.

Issue #857, opened 2017-07-21
[ng] constants for directives not very much mentioned


Janamou:
Hi,

I have written something about this already here: #856.
The thing is, that lately there was a change, that if I need to use some directives - ngIf, ngModel etc., I need to also include directive constants - FORM_DIRECTIVES, CORE_DIRECTIVES etc...

And this is not very much written (it's kind of important information) like a big warning in the documentation (its only in tutorial). I am talking about these articles:

(List updated by @chalin)

- [x] https://webdev.dartlang.org/angular/guide/displaying-data
- [x] https://webdev.dartlang.org/angular/guide/forms
- [x] https://webdev.dartlang.org/angular/guide/router
- [ ] https://webdev.dartlang.org/angular/guide/template-syntax
- [ ] https://webdev.dartlang.org/angular/guide/structural-directives
- [x] ~https://webdev.dartlang.org/angular/guide/user-input~, #1040

kwalrath:
Thanks for the feedback, Jana. You're right that we need to be explicit about the requirements (imports, constants) for each page.

chalin:
How about an alert like the following (most of the text is borrowed from what we used in the toh-1 page):
<img width="838" alt="screen shot 2017-09-08 at 12 15 41" src="https://user-images.githubusercontent.com/4140793/30221001-a9e19bb2-948f-11e7-86a5-bc363ef7b6c3.png">

This particular alert is for the displaying-data page. I would place it towards the end of the first section that uses a core directive.

kwalrath:
I like it, but I'd tighten up the language and emphasize individual directives a bit more. (We plan to encourage individual directives once the tools make that easier.) Maybe:

> ...annotation. You can list directives individually, or for convenience you can use groups like CORE_DIRECTIVES:

/cc @matanlurey

chalin:
@kwalrath - I've implemented your suggestion.

A few more remarks and questions:
- #1040 cleans out the user-input example, removing angular_forms, since it is not needed.
- The [Template Syntax](https://webdev.dartlang.org/angular/guide/template-syntax) page is a collection of disjoint topics that covering directives from the core and forms. Do we really want to add a general comment at the top that reminds folks to add directions (without being specific about which directives)?
- A similar comment applies to the Structural Directives page.

chalin:
#1041 covers all cases except template-syntax and structural-directives.

kwalrath:
[...]

PR #1041, opened 2017-09-08
More prominent mention of angular_* pkg dependencies, directives, etc.

chalin:
Contributes to #857 with prose updates for:

- displaying-data: add CORE_DIRECTIVES to directives list.
- forms: pubspec dependency; page already described need for formsDirective.
- router (two pages): pubspec dependency, imports, providers and directives lists.
- toh-1 wording adjustment

Staged at 
- https://angulardart-org-dev2.firebaseapp.com/angular/guide/displaying-data#ngFor
- https://angulardart-org-dev2.firebaseapp.com/angular/guide/forms#setup
- https://angulardart-org-dev2.firebaseapp.com/angular/guide/router#router-package
- https://angulardart-org-dev2.firebaseapp.com/angular/guide/router/1 (top of page)
- https://angulardart-org-dev2.firebaseapp.com/angular/tutorial/toh-pt1#componentdirectives-


src/angular/tutorial/toh-pt1.md
```diff
@@ -167,9 +167,9 @@ library, it isn't available by default.
 
 Before you can use any Angular directives in a template,
 you need to list them in the `directives` argument of your component's
-`@Component` annotation. Although you can list directives individually,
-it is sometimes more convenient to include a group of related directives, like
-the [form directives][formDirectives], in one go (note the new import statement):
+`@Component` annotation. You can list directives individually, or for
+convenience you can use groups like [form directives][formDirectives]
```
kwalrath: I like the way you said it for ROUTER_DIRECTIVES (plus we should refer to formDirectives by name):

list directives -> add directives
<last line> -> convenience you can add the [formDirectives][] list.

src/angular/guide/router/1.md
```diff
@@ -15,7 +15,17 @@ nextpage:
 {% include_relative _milestone-nav.md  selectedOption="1" %}
 
 Create a new project named `router_example`, and populate it as described in
-[Setup for Development](/angular/guide/setup).
+[Setup for Development](/angular/guide/setup). Then add the router and forms libraries as
```
kwalrath: libraries -> packages

Should we be more precise about the package names in the text?
router -> angular_router, forms -> angular_forms?

src/angular/guide/router/index.md
```diff
@@ -87,6 +87,25 @@ You'll learn many router details in this guide which covers
 * Choosing the "HTML5" or "hash" [URL style](#browser-url-styles)
 {% endcomment %}
 
+## Router package
+
+The [angular_router][] library comes in its own [package][].
```
kwalrath: We haven't mentioned angular_router before in this page. Maybe:

Router functionality is in the [angular_router][] library, which comes in...

its own [package] -> [its own package]
(to better imply the destination of the link)

(for maintainability, it might also be good to use clearer link destination names, since angular_router could reasonably be either the API docs or the package page)

src/angular/guide/router/index.md
```diff
@@ -87,6 +87,25 @@ You'll learn many router details in this guide which covers
 * Choosing the "HTML5" or "hash" [URL style](#browser-url-styles)
 {% endcomment %}
 
+## Router package
```
kwalrath: Router package -> The angular_router library and package
? (since this section covers both)

or maybe:
-> angular_router
-> Adding angular_router

src/angular/guide/forms.md
```diff
@@ -88,8 +88,33 @@ We'll build this form in small steps:
 
 ## Setup
 
-Follow the [setup](setup.html) instructions for creating a new project
-named `angular_forms`.
+Follow the [setup](setup.html) instructions for creating a new project named `forms`.
+
+<?code-excerpt path-base="examples/ng/doc"?>
+
+The `angular_forms` library comes in its own package. Add the package to the pubspec dependencies:
```
kwalrath: This is the first mention of angular_forms on this page, so let's treat it similarly to how we treat angular_router (see my comments, below).

That includes whether we use a diff vs. a plain old pubspec listing.

Similarly to the router page, this should probably be its own section separate from "Setup", called "The angular_forms library and package".

(It's a little strange to have setup in a non-tutorial page, but I don't want to deal with that right now...)

src/angular/tutorial/toh-pt1.md
chalin: Fixed

src/angular/guide/router/1.md
chalin: My thinking was: the shorter names keep the prose a bit lighter, and the code excerpt that follows has the full package names. WDYT?

chalin: > libraries -> packages

Done

src/angular/guide/router/index.md
chalin: > -> Adding angular_router

Good suggestion. Going with that.

src/angular/guide/router/index.md
chalin: All good suggestions. Done

For the link destination names, I've (mostly) consistently used `angular_foo` to refer to the library's API page and `angular_foo@pub`.  Of course, we can change that, but it would be nice to have short names for what is used most often.

src/angular/guide/forms.md
chalin: > This is the first mention of angular_forms on this page, so let's treat it similarly to how we treat angular_router

Done.

> That includes whether we use a diff vs. a plain old pubspec listing.

Since this is part of a "setup" process, IMHO, it makes sense to show a diff. Furthermore, using diffs allows us to avoid adding more docregions.

> Similarly to the router page, this should probably be its own section separate from "Setup", called "The angular_forms library and package".

That would result in a Setup section with only one line:
> **Setup and add angular_forms**
>
> Follow the [setup](setup.html) instructions for creating a new project named `forms`.
>
> **Adding angular_forms**
> ...

How about having only one section entitled:
> **Setup and add angular_forms**

kwalrath: Having setup (and, in general, getting the base example code) in the TOC doesn't seem important here. How about we just focus on adding the package? 

## Add angular_forms

chalin: SGTM

src/angular/guide/router/index.md
chalin: Shall we be consistent with the forms page and change this to:

> **Add angular_router**

kwalrath: I thought about that, but since this is a guide rather than a tutorial, I don't think we're actually telling people to **do** this right then. Make sense?

chalin: Yes, that makes sense. I've changed it back to "Adding ..."

src/angular/guide/forms.md
```diff
@@ -86,10 +86,37 @@ We'll build this form in small steps:
 1. Handle form submission with **ngSubmit**.
 1. Disable the form’s submit button until the form is valid.
 
-## Setup
+## Add angular_forms
 
-Follow the [setup](setup.html) instructions for creating a new project
-named `angular_forms`.
+Follow the [setup](setup.html) instructions for creating a new project named `forms`.
```
kwalrath: for creating -> to create

(Sorry, I missed this one earlier.)

chalin: Np. Done

chalin:
Restaged.

src/angular/tutorial/toh-pt1.md
```diff
@@ -167,9 +167,9 @@ library, it isn't available by default.
 
 Before you can use any Angular directives in a template,
 you need to list them in the `directives` argument of your component's
-`@Component` annotation. Although you can list directives individually,
-it is sometimes more convenient to include a group of related directives, like
-the [form directives][formDirectives], in one go (note the new import statement):
+`@Component` annotation. You can add directives individually, or for
+convenience you can add the [formDirectives][] list
+(note the new import statement):
```
kwalrath: Reading this through, the ordering of the angular_forms library/package discussion is a bit weird.

It says NgModel is... defined in the angular_forms library, but you haven't introduced that library. And you normally wouldn't import the library before adding the package to pubspec.

Fixing this might require some thought, since the point it's trying to illustrate (needing to list things in `directives`) is orthogonal to adding a package. I'll file a bug about this.

PR #1039, opened 2017-09-08
[WIP] chore: re-enable API doc generation for angular_test

chalin:
Fixes #1011

PR #1064, opened 2017-09-21
Support API doc generation for any angular package version including git HEAD

### LINKED ISSUE #1011: API: re-enable inclusion of angular_test API
Just before releasing the 4.0 docs, we  had dartdoc hang (on Travis, not locally) when attempting to generate the API docs over `angular_test`.

To add `angular_test` back, simply drop the package list given on the command line, here: https://github.com/dart-lang/site-webdev/blob/master/.travis.yml#L23

chalin:
Fixes #1011
Supersedes #1039


kwalrath:
It's good to see the angular_test API docs!

chalin:
Yes! And now we can generate the router 2.x API from GH too!

Issue #601, opened 2017-05-05
Confusion in Master/Detail tutorial


LindaLawton:
I will be referring to this page. [Master/Detail](https://webdev.dartlang.org/angular/tutorial/toh-pt2)

There are several places in the tutorial where the code location is pointed to `lib/app_component.html (hero list property)`  The issue i have with this is that at this point in the series we have not gone into creating app_components.html and at the end of the tutorial all of the html is placed in @Component(    template: ' ') as a beginner to Dart this caused me a few minutes of confusion but as an expected developer I was able to figure out the problem quickly. 

Options:

 - Either remove the  `lib/app_component.html (hero list property)`  and explain it differently.
 - Show the reader how to create and use app_component.html  at the beginning of the tutorial.  

Personally I like the second option I think having the that much html in the template parm is messy.

kwalrath:
Agreed. It seems like we should have separate .css & .html files from the start. @chalin, is there a good reason (any more) to have everything in the .dart file?

LindaLawton:
Give me a few days to go through the series I may have more comments. No reason for you guys to go through it more than once. 

I haven't done much web development in the last five years so I am doing this from the complete novice state.

kwalrath:
OK, we'll hold off on changes for now. Thanks for taking the time to give us feedback!

LindaLawton:
Defines AppComponent tests. While testing isn't covered in this tutorial, you can learn how to test the Tour of Heroes app from the Testing page.

If you aren't covering tests in the tutorial and they aren't required. Why even include this?

LindaLawton:
[Multiple Components](https://webdev.dartlang.org/angular/tutorial/toh-pt3)   I think the order you are explaining things here is a little confusing.   

1. create hero_detail_component.dart 
2. Add hero property
3. create hero.dart.
4. import hero.dart
5. Add the template and change selectedHero to hero 

>The hero property is an input

Its a little confusing where this needs to go but i think that will be fixed when you add the .html file and get all the template stuff out of the dart file.

LindaLawton:
[Services](https://webdev.dartlang.org/angular/tutorial/toh-pt4)
>Inject the HeroService
Three lines replace the one line of new:

I am only seeing two lines  
```
final HeroService _heroService;
AppComponent(this._heroService);
```

[...]

Issue #1015, opened 2017-09-03
Change default of `<code-pane>` to no linenums


chalin:
Currently, code excerpts in a `<code-pane>` show linenums by default. This was chosen as a default because most cases used linenums, but we are getting enough cases now that should _not_ have linenums that it is worth switching the default (as opposed to introducing, say, a `no-linenums` option).

PR #1082, opened 2017-09-28
As of toh-2, move app template to its own file

chalin:
- Initially requested as part of #601.
- This change will also help the upcoming edits to toh-5 which use the new router.

Staged at https://angulardart-org-dev2.firebaseapp.com/angular/tutorial/toh-pt2, etc.


kwalrath:
Found the details!

The example that seemed old was in https://angulardart-org-dev2.firebaseapp.com/angular/tutorial/toh-pt4. lib/app_component.dart has a `template` field, instead of `templateUrl`.

The one-liner with a line # was in https://angulardart-org-dev2.firebaseapp.com/angular/tutorial/toh-pt5#style-the-navigation-links, just after the "Style the navigation links" header.

chalin:
> Found the details!

Great!

> The example that seemed old was in https://angulardart-org-dev2.firebaseapp.com/angular/tutorial/toh-pt4. ...

I believe that that was just a glitch (maybe I didn't to a full site regeneration before deploying). That file is correctly updated in my local copy. I've redeployed and the [staged copy is fine too](https://angulardart-org-dev2.firebaseapp.com/angular/tutorial/toh-pt4#review-the-app-structure). (If you are still seeing the template, maybe empty your cache.)

> The one-liner with a line # was in https://angulardart-org-dev2.firebaseapp.com/angular/tutorial/toh-pt5#style-the-navigation-links, just after the "Style the navigation links" header.

That will be fixed as part of #1015 (it can't really be fixed in a one-off fashion).

> I like the way you've compressed the directory structure lists, using {css, dart, html}. Much easier to read.

:+1:

So I don't think that there are any further changes to be made to this PR.

kwalrath:
The code looks good now.

PR #1083, opened 2017-09-28
Move version and changelog pages to top level

### LINKED ISSUE #1002: Move version and changelog pages to top level
@kwalrath - WDYT?

chalin:
Fixes #1002

(Not staged since I'm running out of staging servers :)


src/changelog.md
```diff
@@ -1,10 +1,8 @@
 ---
-layout: angular
 title: AngularDart Changelog
 description: A summary of changes to the AngularDart documentation and examples.
```
kwalrath: AngularDart -> site

src/changelog.md
```diff
@@ -1,10 +1,8 @@
 ---
-layout: angular
 title: AngularDart Changelog
```
kwalrath: Delete "AngularDart"?

chalin: done

src/changelog.md
chalin: done

firebase.json
```diff
@@ -5,7 +5,9 @@
     "trailingSlash": false,
     "redirects": [
       { "source": "/articles/low-level-html/improving-the-dom", "destination": "/guides/web-programming", "type": 301 },
+      { "source": "/angular/changelog", "destination": "/changelog.html", "type": 301 },
```
kwalrath: Why do the destinations have ".html" suffixes?

chalin: IIRC, it doesn't work w/o the `.html` suffix. The target is, say, `foo.html`. The server strips the `.html` because of `clearUrls: true`.

kwalrath: That's weird, it doesn't work that way on site-www.

chalin: I've double checked and the redirects work (at least locally ... they should also work on webdev).

PR #1086, opened 2017-09-29
chore: log footer build info

chalin:


chalin:
The Travis `build` task is passing, which is what this PR is meant to help diagnose (when the task is run on master, that is).

PR #1087, opened 2017-09-29
chore: footer build info tweaks

chalin:
Make clear that the build info is for the site build, not angular builds.
Also ensure that the build timestamp is set.


kwalrath:
We should probably add tooltips for the other footer icons, too. Maybe:

* This site's source is on GitHub.
* File an issue about this site.
* Site build: 2328 (2017/09/29 16:00 PDT)

chalin:
Oops, I just saw your comment now. I'll followup via #1088.

PR #1108, opened 2017-10-13
Changelog entry covering switch to HashLocationStrategy

chalin:


src/changelog.md
```diff
@@ -27,6 +27,19 @@ but you can see a preview at
 </aside>
 {% endcomment %}
 
+## Router _HashLocationStrategy_ (October 2017)
+
+Switched to using [HashLocationStrategy][] in the
+[Router][] example and the [Tutorial][]. In this way,
+features like deep linking into the example apps work as expected.
```
kwalrath: A little more explanation might help here. Maybe something like this:

...as expected when no server-side support is available.
OR
...as expected during development.

Issue #1054, opened 2017-09-17
Multiple # in links causing issues


ninachen:
_Example workflow_
I would like to share a link to a particular section in the docs, such as
https://webdev.dartlang.org/angular/guide/displaying-data#!#using-the-hero-class

If I try to send this link through gmail, gmail will convert it to
https://webdev.dartlang.org/angular/guide/displaying-data#!/%23using-the-hero-class
which doesn't resolve correctly and just links to the top of the page.

_Expected result_
I believe that the best solution would be for the doc URL to only have one #, as it appears to be a [violation of the spec](https://tools.ietf.org/html/rfc3986#section-3.5) to have multiple # characters in a fragment identifier. Given that, the existing gmail behavior might be intentional? (Interestingly, github appears to be okay with this.)

I am unaware of what other use cases might be affected: happen to be using gmail here as an illustrative example.

So instead of ending in `#!#using-the-hero-class`, the URL should end in `#using-the-hero-class`:
https://webdev.dartlang.org/angular/guide/displaying-data#using-the-hero-class

Also, using `#foo` autoresolves to `#!#foo`, so at least that gives me a workaround for the time being.

PR #1109, opened 2017-10-13
Fix a link for built in pipes

camsteffen:


chalin:
Follow-through of #1084, contributing to the fix for #1054.

PR #1118, opened 2017-10-16
Versions

kwalrath:
I was having a hard time reading the package names in the table, so I reformatted it. (It could use more work, but I think this is an OK start.) Also, I added a request for issue reports to the alpha section.

Staged at: https://kw-webdev-dartlang-1.firebaseapp.com/version

Ideas for more work to do (not necessarily in this PR):

* Redirect/change versions to version (I keep typing in the wrong URL).
* Make it clearer what the links in the table go to.


chalin:
LGTM too!

Issue #1065, opened 2017-09-21
[5-dev] Prepare webdev, and all angular-example repos for 5-dev


chalin:
Once a first 5-dev release is available:
- [x] Create a `5-dev` branch.
- [x] Setup autodeploy of 5-dev to dev staging server (this was actually still part of the base setup).
- [x] Update versions page on master - #1133
- [x] Update versions page on 5-dev - #1130
- [x] Adapt build scripts so that
  - 5-dev only builds over the dev channel  - #1129, #1130
  - master only builds under stable - #1128, #1138

PR #1133, opened 2017-10-24
Add next versions of packages and fix version page generation

chalin:
(Version page was using `site.next-url` instead of `site.dev-url`)

Contributes to #1065

Staged at https://angulardart-org-dev2.firebaseapp.com/version
Contrast with: https://webdev.dartlang.org/version

---
This is what the new versions table looks like:
screen shot 2017-10-24 at 19 17 14


Sfshaza:
I trust you've tested, because I'm not sure how this versioning works. If you have, then LGTM.

chalin:
Tested, and Travis reports green.

PR #1144, opened 2017-10-26
Make Angular example component tests less brittle

chalin:
Contributes to #1065


examples/ng/doc/architecture/lib/src/sales_tax_component.dart
```diff
@@ -10,7 +10,7 @@ import 'tax_rate_service.dart';
       Amount: <input #amountBox (change)="0">
 
       <div *ngIf="amountBox.value != ''">
-      The sales tax is
+      The sales tax is &ngsp;
```
kwalrath: ngsp -> nbsp?

chalin: [`&ngsp;` is an Angular space.](https://github.com/dart-lang/angular/blob/6fc43bb170683432f4532877d54abb3ccbc676c6/angular/lib/src/core/metadata.dart#L320).

`preserveWhitespace` was introduced back in Angular 2, but the default was flipped in Angular 5:  ["preserveWhitespace is now false by default in @Component".](https://github.com/dart-lang/angular/blob/master/angular/CHANGELOG.md#breaking-changes-1). 

This article has some details: http://blog.ninja-squad.com/2017/09/18/what-is-new-angular-4.4.

I've created #1166 to follow through with adding some documentation about this.

kwalrath: ah ok. I was wondering if it was an ng-inspired typo or an actual thing.

PR #1176, opened 2017-11-15
Improve print styles

### LINKED ISSUE #35: Fix print styles
See https://github.com/dart-lang/site-www/issues/149 for details.

undecidabot:
Fixes #35.

This PR contains the following style changes when printed:

- Puts solid black borders around notes and code blocks
- Puts an underline under links (and removes the path or external link icon)
- Shows all of the code panes instead of just the selected tab


kwalrath:
I'm going to take a closer look in a bit, when I have access to a bigger screen and a printer, but at first glance, this looks great! Being able to see all the code panes is a huge improvement.

src/_assets/stylesheets/_overwrites.scss
```diff
@@ -120,12 +120,66 @@ img.centered {
...
+  a:after {
+    content: none !important;
+  }
+
+  /* Get rid of the space for the icon */
+  a.external {
+    padding-right: 0;
+  }
+
+  /* Hide back to top links */
```
kwalrath: Nice!

src/_assets/stylesheets/_overwrites.scss
```diff
@@ -120,12 +120,66 @@ img.centered {
...
+    display: none;
+  }
+  
+  /* Show all panes in the code box */
+  code-pane > .ng-hide:not(.ng-hide-animate) {
+    display: block !important;
+  }
+
+  /* Add the file path for the code panes mimicking the style of the code-example header text */
+  code-pane:before {
```
kwalrath: Could you do it after? When you print the pages, often there's a page break between the file path and the code pane.

kwalrath: Hmmm... I just mocked up after, locally, and it doesn't look good to me.

Is there a way we could have the path before (or in, to the top right?) but not allow a page break between the path and the code pane?

undecidabot: Yeah, I did notice this issue as well. I placed it on top to follow the other code examples. I tried using a CSS property to avoid page breaks but it didn't seem to work. Although we can make it look like it's inside of the "code box", since it's only visually and not actually inside, the page breaks can still split the two part.

undecidabot: I just investigated this again and removing the `page-break-inside: avoid` of the `pre` tag makes the splits a lot less likely with the side effect of having more code blocks split between pages and a few pages saved. We could add that property to the `code-pane` and `.code-example` instead to avoid splits between the path and the code block. What do you think?

kwalrath:
I tried adding this to the print styles, and it seems to do the right thing:

```
+  /* Save space and avoid page breaks after file paths. */
+  pre {
+    page-break-inside: auto;
+  }
```

So I'm going to accept your change, and then do a subsequent PR with this fix.

Thanks so much for your CSS fixes and advice!

Issue #737, opened 2017-06-14
[ng] selected chapters need to be copyedited/synced


chalin:
Some chapters still use the form of "our/we". Sync it with the corresponding TS page, which has been revised by tech writers.
- [x] dependency-injection - #1182
- user-input
  - [x] #1179 
  - [x] #1184

chalin:
All done in time for the WtD sprint! 🎉

PR #1179, opened 2017-11-15
Sync of angular user-input page

chalin:
Contributes to #737 

Staged at https://angulardart-org-dev2.firebaseapp.com/angular/guide/user-input


src/angular/guide/user-input.md
```diff
@@ -311,49 +335,24 @@ We'll call out the highlights after we bask briefly in its minimalist glory.
...
-
-Although the example *works*, we are rightly wary of Dart in HTML.
-Template statements are powerful. We're supposed to use them responsibly.
-Complex Dart in HTML is irresponsible.
-
-Should we reconsider our reluctance to pass the input box into the component?
-
-There should be a better third way. And there is, as we'll see when we learn about `NgModel` in the [Forms](forms.html) chapter.
+- **Keep template statements simple** &mdash; The `(blur)` event is bound to
+  two JavaScript statements. The first statement calls `addHero()`. The second
```
kwalrath: Are these really JavaScript statements?

kwalrath: Also the explanation doesn't match the observation. Looking at the old version, the whole point of this observation was to say that the blur binding was too complicated.

**Keep template statements simple.** Template statements should be simple method calls or property accesses. The template statement for the `(blur)` event performs two method calls, which is questionable. There's a better way to [PENDING: summarize what this code is trying to accomplish], as the Forms page explains.

src/angular/guide/user-input.md
```diff
@@ -364,10 +363,14 @@ Here is all the code we talked about in this chapter.
 
 ## Summary
 
-We've mastered the basic primitives for responding to user input and gestures.
-As powerful as these primitives are, they are a bit clumsy for handling
-large amounts of user input. We're operating down at the low level of events when
-we should be writing two-way bindings between data entry fields and model properties.
+You have seen the basic primitives for responding to user input and gestures.
```
kwalrath: You have -> You've

src/angular/guide/user-input.md
```diff
@@ -10,40 +10,41 @@ nextpage:
...
 ## Binding to user input events
 
-We can use [Angular event bindings](./template-syntax.html#event-binding)
-to respond to [any DOM event](https://developer.mozilla.org/en-US/docs/Web/Events).
+You can use [Angular event bindings](template-syntax#event-binding) to respond
+to any [DOM event][]. Many DOM events are triggered by user input. Binding to
+these events provides a way to get input from the user.
+
+To bind to a DOM event, surround the DOM event name in parentheses and assign a
+quoted [template statement](template-syntax#template-statements) to it.
```
kwalrath: Delete this blank line. There's no need for a separate paragraph here.

src/angular/guide/user-input.md
```diff
@@ -99,31 +100,51 @@ The user data we want is in that variable somewhere.
...
+All standard DOM [Event][] objects have a `target` property, a reference to
+the element that raised the event. In this case, `target` refers to the
+[`<input>` element][HTMLInputElement] and `event.target.value` returns the
+current contents of that element.
 
-Enter the letters "abc", and then backspace to remove them.
-Here's what the UI displays:
+After each call, the `onKey()` method appends the contents of the input box
+value to the list in the component's `values` property, followed by a
+separator character (|). The [interpolation](template-syntax#interpolation)
```
kwalrath: "Interpolation" doesn't itself display anything, and we don't want people to confuse this with Dart interpolation. Maybe something like this:

The template uses Angular interpolation (`{{...}}`) to display the `values` property.

src/angular/guide/user-input.md
```diff
@@ -137,23 +158,32 @@ Here's what the UI displays:
...
+`InputElement` &mdash; not all elements have a `value` property. The `onKey()`
+method more clearly expresses what it expects from the template and how it
+interprets the event.
 
-  We'll address this problem in our next try at processing user keystrokes.
-</div>
+### Passing _$event_ is a dubious practice
 
+Typing the event object reveals a significant issue with passing the entire
+DOM event into the method: the component has too much awareness of the
```
kwalrath: We should remove the [anthropomorphism](https://developers.google.com/style/anthropomorphism). Software isn't (yet!) aware or knowledgable.

has too much awareness of -> is closely tied to

src/angular/guide/user-input.md
```diff
@@ -137,23 +158,32 @@ Here's what the UI displays:
...
+template details. It can't extract information without knowing more than it
```
kwalrath: It can't.... -> The component can't extract data without using web APIs.

src/angular/guide/user-input.md
```diff
@@ -78,14 +82,11 @@ This time we'll (1) listen to an event and (2) grab the user's input.
...
 <div class="callout is-important" markdown="1">
   <header> $event vs. \$event</header>
-  Templates in Dart files need a `\` in front of the `$`.
-  If the template is in an HTML file, use `$event` instead of `\$event`.
+  Non-[raw strings]({{site.dartlang}}/guides/language/language-tour#strings) in
+  Dart files need a `\` in front of the `$`. If the template is in an HTML
+  file, use `$event` instead of `\$event`.
 </div>
 
 <?code-excerpt "lib/src/keyup_components.dart (class v1)" region="key-up-component-1-class-no-type" title?>
```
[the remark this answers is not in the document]
kwalrath: Please address this one.

src/angular/guide/user-input.md
kwalrath: Please address this comment, since it requires knowledge of Angular.

src/angular/guide/user-input.md
```diff
@@ -99,31 +100,51 @@ The user data we want is in that variable somewhere.
...
 <div class="l-sub-section" markdown="1">
-  We declare the `$event` as a `dynamic` type, which means we've abandoned strong typing
-  to simplify our code. We generally prefer the strong typing that Dart affords.
-  We can rewrite the method, casting to HTML DOM objects like this.
+  Alternatively, you could accumulate the individual keys themselves by
+  substituting `event.key` for `event.target.value` in which case the same user
+  input would produce:
+
+  ```nocode
+    a | b | c | backspace | backspace | backspace |
```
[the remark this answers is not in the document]
kwalrath: Please address this comment.

src/angular/guide/user-input.md
kwalrath: Please address this one, since it requires subject matter knowledge.

src/angular/guide/user-input.md
kwalrath: Please address these comments.

chalin:
Yes, SGTM :)

src/angular/guide/user-input.md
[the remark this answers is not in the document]
chalin: Code excerpt moved.

> is there a good reason to show anything more than the onKey method in the second excerpt? (The classname is distracting.)

It annoyed me too that this page has class names suffixed with a version (`V1`, `V2`, ...), but I chose not to address that during this edit cycle.

For the code excerpt, since this is the first time that this particular class is discussed, I think that it would be better to show the entire (small) class body.

src/angular/guide/user-input.md
chalin: Yes, much better (that statement had bothered me too).

Fixed (with a few more tweaks to the preceding sentence).

src/angular/guide/user-input.md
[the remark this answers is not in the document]
chalin: (Actually, all of the different versions of the keyup component are show in the app, so it will be easier to leave the version suffix on the component class names.)

src/angular/guide/user-input.md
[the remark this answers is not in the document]
chalin: Good catch. Fixed.

src/angular/guide/user-input.md
chalin: Done.

src/angular/guide/user-input.md
chalin: > Are these really JavaScript statements?

Oops, no, that slipped through (they are Angular template statements, as you point out). Fixed.

> Also the explanation doesn't match the observation.

I feel that it does. Admittedly, `(blur)` is bound to _two_ such simple statements, which contributes to a code smell. But being a code smell, doesn't make it outright unacceptable.

I feel that the revised observation prose given in the PR "separates concerns", relative to original prose. I.e., the observation "Keep template statements simple" is relevant and is kept positive. In the summary that follows immediately after, we use the "there is a better way" as a segue to the next page.

I'd vote to keep it that way. WDYT?

src/angular/guide/user-input.md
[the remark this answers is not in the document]
chalin: Come to think of it, the proper way to name these component classes would be, KeyUpXComponent (where X is the variant qualifier). I'll make that change.

chalin:
Code cleanup commit submitted.

PR #1188, opened 2017-11-17
Document functional attribute directives

chalin:
Contributes to #1111

Note that there are also functional _structural_ directives, which will be documented in the section on "Structural directives".

Staged at https://angulardart-org-dev.firebaseapp.com/angular/guide/attribute-directives

PR #1189, opened 2017-11-18
Functional attribute directive example: use `@Attribute()`, add E2E

chalin:
- Enhance example: illustrate use of `@Attribute()`.
- Add E2E test for new functional attribute directive.
- Upgrade E2E to use async/await.

Followup to #1188


chalin:
@leonsenft - if you have a minute, could you look at the small functional attribute directive example and let me know if you have any comments? Thanks.

leonsenft:
@chalin Looks good to me! If you were concerned about testing this, you could inject an instance of a counter object instead of using the global integer (imagine you had 100 tests and you removed one in the middle, all the following numbers would be off), but for a simple example like this I think that's needless complexity.

PR #1220, opened 2017-12-21
Copy edit: angular/guide/template-syntax (PR 1200 followup)

sarahmaddox:
Includes updates from https://github.com/dart-lang/site-webdev/pull/1200 (@100percentlinen) and from comments in that PR.

PR #1224, opened 2017-12-27
Tweaks to template-syntax

kwalrath:
Follow-up to #1220 

Staged at https://kw-webdev-dartlang-2.firebaseapp.com/angular/guide/template-syntax

(Btw, that deploy issue showed up again, so I added to the internal bug.)

I plan to commit this, because I think it might unbreak the build.


kwalrath:
Indeed, this does fix the build.

PR #1201, opened 2017-11-24
Copy edit: angular/guide/testing/component

jenapp:
Still possible issues with the formatting of the package list at the top, but out of time.

PR #1225, opened 2017-12-27
Fix formatting in component testing index

kwalrath:
Followup to #1201


jenapp:
Thanks for sorting the HTML-markdown issue Kathy! I'll keep it in mind in the future.

PR #1285, opened 2018-01-24
[5-dev] reinstate dart2js-based testing

### LINKED ISSUE #1284: [5-dev] Resume test, build and deployment using dart2js

chalin:
Fixes #1284

Component tests are run under dartdevc only (for those that _are_ running, which is only two at the moment). All active E2E tests are run under both compilers (pending example migration to Angular 5 and bugs in dart2js).


chalin:
All jobs are green:
screen shot 2018-01-24 at 18 09 26

PR #1299, opened 2018-01-31
[5-dev] Changelog update, and added build_* pkg version freshness check

chalin:


chalin:
@kwalrath gave a 👍 offline.

PR #1324, opened 2018-02-13
[5-dev] Use Travis stages

chalin:
- Builds used to last >20 min, for a total time of 2.5 hours; e.g. see [this Travis log](https://travis-ci.org/dart-lang/site-webdev/builds/339555931).
- Once caches are updated, the current build requires <20 min for a total time of <50 min; e.g. see [this Travis log](https://travis-ci.org/dart-lang/site-webdev/builds/341190078).

Almost all of the benefits are from improved caching, not stages. There is room for further improvements, but this will do for now, while we move on to other work items.

(For some reason fonts are not being generated in woff2 format. Temporarily added rule to link check skip list.)


jakemac53:
Note that the travis cache is keyed off of the environment variables and dart sdk version, so currently most of these tasks do not share a cache with each other. That might be desired or not, I don't know enough about the builds to know its worth mentioning :).

To get around that in mono_repo we pass the task as a command line arg instead of using an environment variable.

chalin:
> currently most of these tasks do not share a cache with each other. That might be desired or not ...

@jakemac53 - thanks for double checking, but yes, for the moment, that is the desired behavior. Stages buy us very little in this repo (again, at the moment), except for the main site build. Every other job has its own cache.

Later, once I can re-enable most component tests (that currently don't work because of pageobjects), it might make sense to have a separate example app build stage (with a cache shared by `test` and `e2e` tasks).

cc @kwalrath

jakemac53:
sounds good

Issue #980, opened 2017-08-29
regularize angular & dartlang styles


kwalrath:
Because the Angular docs were brought over from angular.io, they use different styles than the rest of the dartlang.org pages. We should avoid unnecessary differences. For example, look at these two pages:

* https://webdev.dartlang.org/angular/guide/router (angular style)
* https://webdev.dartlang.org/tools/dartdevc (dartlang.org style)

The sidenavs are different colors, the headings are different sizes, and the note boxes have different borders and fonts. I rather like many of the angular style changes, but if we keep them, we should use them throughout [webdev.dartlang.org](https://webdev.dartlang.org) & www.dartlang.org.

angular style:
![image](https://user-images.githubusercontent.com/2164483/29839353-ca794fda-8cb2-11e7-85ba-dfed8cd8d529.png)

dartlang.org style:
![image](https://user-images.githubusercontent.com/2164483/29839346-c3fcbdea-8cb2-11e7-92b1-342d7c09d963.png)

chalin:
Bootstrap 4 (stable) has been released! https://getbootstrap.com

Related: https://github.com/dart-lang/site-www/issues/527

chalin:
FYI, I've just cleaned out the last few uses of `l-main-section` etc.

PR #1348, opened 2018-02-28
Fix alert styles across webdev pages; replace callout by alert

chalin:
Because Angular pages were adopted from angular.io, some of the page styles differ for Angular pages as compared to the rest of the site. This PR is a first step in consolidating page styles (#980). In particular this fixes the appearance of alerts in page banners.

The callout style has been dropped in favor us using the (more standard) alert style.

There is essentially no content change in this PR.

Staged at https://angulardart-org-dev.firebaseapp.com

cc @kwalrath 

---

Before this PR, top-level page alerts looked like this:

> screen shot 2018-02-28 at 12 36 20

They'll now look like this:

> screen shot 2018-02-28 at 12 35 12


src/angular/guide/dependency-injection.md
```diff
@@ -597,8 +597,9 @@ is a subtype of [Injectable][].  It is in
 fact `Injectable` annotations that
 identify a class as a target for instantiation by an injector.
 
-<div class="callout is-critical" markdown="1">
-  <header> Always include the parentheses</header>
+<div class="alert alert-warning" markdown="1">
+  <h4>Always include the parentheses</h4>
```
kwalrath: The h4 looks a little weird to me. We usually use inline bold text instead of a heading for an aside.

src/angular/guide/pipes.md
```diff
@@ -621,7 +622,7 @@ A breakpoint on the pipe's request for data shows the following:
 In the previous code sample, the second `fetch` pipe binding demonstrates more pipe chaining.
 It displays the same hero data in JSON format by chaining through to the built-in `JsonPipe`.
 
-<div class="callout is-helpful" markdown="1">
+<div class="alert alert-info" markdown="1">
   <header> Debugging with the json pipe</header>
```
kwalrath: Instead of a header, this should be **Debugging with JsonPipe:**

kwalrath:
I'd argue that the precedent wasn't there before we brought over the Angular docs... It's strictly an Angular convention.

But I'm fine with fixing it later.

chalin:
Right, that did come from Angular docs.

> But I'm fine with fixing it later.

Here is the issue to track that: #1349

PR #1353, opened 2018-03-02
Fix router code sample

pd4d10:
Add directives and providers to make the router code sample work


chalin:
Thanks for the PR, but the code change you are suggesting isn't valid: the [Route](https://webdev.dartlang.org/api/angular_router/angular_router/Route-class) constructor does not have `directives` and `providers` as named parameters.

The section on [Declaring router providers and directives](https://webdev.dartlang.org/angular/guide/router/1#declaring-router-providers-and-directives) explains how it should be done.

pd4d10:
@chalin My mistake. It should be placed at `@Component` decorators 😅

Specifying `ROUTER_PROVIDERS` in bootstrap function brings router providers to all components without declaring it one by one. I missed this section. Thanks!

chalin:
Actually, the overview section should be clearer about adding directives and providers. I've created #1356 to track that. Thanks for having pointed that out!

PR #1355, opened 2018-03-02
[5-dev] DI page: update bootstrapStatic description

chalin:


src/angular/guide/dependency-injection.md
```diff
@@ -359,18 +359,26 @@ Applications are bootstrapped in `web/main.dart`:
...
   void main() {
-    bootstrap(AppComponent);
+    bootstrapStatic(AppComponent, [], ng.initReflector);
   }
 ```
 
 The first argument to `bootstrap()` is the app root component class.
-The second optional argument is a providers list. For example:
+The second argument is a providers list.
+The last argument is the statically generated provider registration function
```
kwalrath: This makes me wonder how it's generated and where I can find more information. I'd expect at least a link to more info about main.template.dart and initReflector.

kwalrath: If this is going away, we should say so... That way we can get away with less explanation if it's a hack.

chalin:
@kwalrath I've added the note that you suggested. If/when we do more fully describe generated files, it should probably be in another section. Until the dust settles, might we consider this minimal change enough of a step in the right direction?

PR #1368, opened 2018-03-03
chore(README): update env-set.sh and site installation instructions

chalin:
The `env-set.sh` script now runs nvm and rvm `install` commands by default.


kwalrath:
Thanks! A couple of weeks ago I had some issues with getting Ruby to update, and I kept meaning to add something to the README about that...

PR #1377, opened 2018-03-05
Update external link in docs - user data dir

camsteffen:


chalin:
Thanks for catching that and submitting a PR!

Issue #1366, opened 2018-03-02
[5-dev] Site changes due to upgrade to Dart 2 SDK + new tooling


chalin:
Below is a list of pages that will need to be updated due to changes in the Dart 2 SDK and associated tooling. (This is independent of updates necessary because of upgrades to Angular 5.)

- Angular
  - deployment.md: e.g., mentions `dart_to_js_script_rewriter`, `pub build` etc.
- Tools
  - Drop Dartium, (add Dart Code?)
  - Subpages: dart2js, ddc, pub build & serve
- Articles
  - low-level-html/embedding-in-html.md: e.g., mentions `browser` package
- FAQ

kwalrath:
Thanks for starting this list!

kwalrath:
Of course, we'll need to update anything that mentions core API that has changed (e.g. due to constants being changed from SCREAMING_CAPS to lowerCamelCase). I'm hoping that our tests will catch most or all of these.

PR #1397, opened 2018-03-14
[5-dev] pub build/serve --> build_runner

### LINKED ISSUE #1393: Create a `pub build/server` --> build_runner migration guide
Related to the other work..

chalin:
This is a first quick fix to replacing the `pub build` and `pub serve` entries by a link to the build_runner docs. We can be more inclusive of a local description later.

Also:
- Dropped the dartium page, removing any references to it.
- Added firebase redirect rules: links to dropped pages land back on the tools page.

Contributes to #1366
Contributes to / fixes #1393 (?)

Staged at https://angulardart-org-dev.firebaseapp.com/tools

cc @kwalrath


firebase.json
```diff
@@ -32,7 +32,11 @@
       { "source": "/codelabs/ng2", "destination": "/codelabs", "type": 301 },
       { "source": "/codelabs/ng2/**", "destination": "/codelabs", "type": 301 },
 
-      { "source": "/angular/guide/router/**", "destination": "/angular/guide/router", "type": 301 }
+      { "source": "/angular/guide/router/**", "destination": "/angular/guide/router", "type": 301 },
+
+      { "source": "/tools/dartium", "destination": "/tools", "type": 301 },
```
kevmoo: These should all redirect to a migration page (TBD). Folks clicking here will have zero context why the pages they are visiting have disappeared.

chalin: Actually, I could replace each of those pages with a comment (specific to the page), that says what the alternative is. Come to think of it, that would be much better than a redirect, since we could then direct readers to a specific anchor on a (migration) page.

kevmoo: 👍

chalin:
@kevmoo - please see #1409

PR #1401, opened 2018-03-15
[5-dev] Dart 2 migration guide

chalin:
A possible next iteration of the migration guide. Added:
- Sections for:
  - Code changes
  - Tool changes, and
  - **Cheatsheet**
- "Migration" entry to sidenav
- Page TOC

Staged at https://angulardart-org-dev2.firebaseapp.com/guides/migration

Note that this PR is over the **5-dev branch**, not master.
Contributes to #1366.


kwalrath:
I'm not sure about the URL. This guide seems to be truly about 1.x -> 2, so that should probably be in its name. Will we want to have an angular v4 -> v5 migration guide, as well? We wouldn't want to lose past versions' migration guides, would we? (I suppose we could just make this page an index into migration guides... but it'd be good to always be able to point to the 1.x -> 2 guide.)

URL aside, here are my thoughts about the page:

* The Cheatsheet is very handy and should probably be higher up (and, with a little intro text, replace what's currently in the "Tool changes" section.
* Dart 1 -> Dart 1.x
* As I mentioned in chat, I'd like to see diffs for pubspec.yaml & web/index.html. It might be in addition to the list... hard to know until I see the diffs.
* The resource list at the top should probably go at the end in a "Resources" section.
* We should have a short summary at the top about what they'll need to change and why. This summary could link to individual sections (maybe).

chalin:
I've done all the changes you suggested except adding diffs.

Those diffs will take a lot of page real estate. How about if we simply provided a link to full project diff like this one for quickstart: https://github.com/angular-examples/quickstart/compare/master...5-dev

chalin:
Links to diffs added. I'm closing this PR and opening one with the changes applied to the /dart-2 page on master. We'll then rebase 5-dev and have copies of the pages under both sites.

chalin:
The `master` version of this PR is #1403.

PR #1403, opened 2018-03-15
Dart 2 migration guide updates

chalin:
Supersedes #1401; the changes requested on that PR have been made here.

Contributes to #1366 

Staged at https://angulardart-org-dev2.firebaseapp.com/dart-2

cc @kevmoo


src/dart-2.md
```diff
@@ -2,34 +2,99 @@
...
+- [Tour of Heroes, part 5][angular-examples/toh-5]
+
+### `pubspec.yaml` {#pubspec-yaml}
+
+- Add new `dev_dependencies`:
+  - `build_runner: {{site.data.pubspec.dev_dependencies.build_runner}}`
+  - `build_test: {{site.data.pubspec.dev_dependencies.build_test}}`, if you are running tests
+  - `build_web_compilers: {{site.data.pubspec.dev_dependencies.build_web_compilers}}`
+- Drop `dev_dependencies`:
+  - <del>`browser`</del>
```
kevmoo: The strike-through rendering is a bit much...

src/dart-2.md
```diff
@@ -2,34 +2,99 @@
...
+- Drop all `transformers`:
+  - <del>`angular`</del>
+  - <del>`dart_to_js_script_rewriter`</del>
+  - <del>`test/pub_serve`</del>
+
+For example, here is a diff of
+[angular-examples/quickstart/pubspec.yaml][]
+with these changes applied.
+
+### `web/index.html` {#web-index-html}
```
kevmoo: This could be `example/sample1.html` too

Be a bit more generic?

src/dart-2.md
chalin: I'd like to keep some visual cue distinguishing deleted elements from updated or added elements. How about something like this:
> screen shot 2018-03-15 at 16 30 40

If not, I can drop the \<del> wrapper entirely.

src/dart-2.md
chalin: How about if I keep `web/index.hml` but add a comment to the effect that similar changes will be necessary for any HTML file that has a \<script> element with Dart file as a `src`?

chalin:
@kevmoo - once the nits are resolved, will this be good to go?

src/dart-2.md
kevmoo: Sure...

src/dart-2.md
kevmoo: that's better...

kevmoo:
> @kevmoo - once the nits are resolved, will this be good to go?

Yup!

kevmoo:
...well, let's see what @kwalrath thinks

src/dart-2.md
chalin: I'm open to other suggestions. It's just that `web/index.html` is more likely to catch the eye than, say, `example/sample1.html` IMHO.

kevmoo:
I was thinking something more general as a header "HTML script" tag or
something

Just be clear that it's not JUST for web/index.html

chalin:
I refreshed the staging server.

PR #1417, opened 2018-03-21
[5-dev] Simplifying Router example (refactoring)

chalin:


chalin:
I've reverted to "Crisis center" in #1419 and/or #1421.

Issue #1369, opened 2018-03-03
[Meta] Site changes due to upgrade to Angular 5 and Dart 2


chalin:
This is a meta issue used to track the completion status of the 5-dev branch. Items are in no particular order at the moment, and mainly list major work items, not minutia.

- [x] Updates due to use of Dart 2 are tracked using #1366 
- [x] Tutorial
  - [x] Setup: explaining new way to serve apps
  - [x] Parts 5 and 6: use **router v2** - #361, #1058
- [x] Guides
  - [x] Router, due to v2 API rewrite - #1058
- [x] Component testing
  - [x] Explain new way to run tests
  - [x] Temporarily use mock pageloader until pageloader v3 is released - #1350
  - [x] Update to pageloader v3 - #1351
- [x] [angular-examples](https://github.com/angular-examples) repos:
  - [x] README (#1431)
  - [x] gh-pages with updated app
- [x] WebStorm - #1582
  - [x] Update screenshots
  - [x] Adjust warning to mention min version
- [x] Remove unnecessary new/const? - #1650
  Waiting on https://github.com/dart-lang/angular/issues/1313

Also:

- Pages to review before Angular 5 final #1422

Issue #1058, opened 2017-09-19
[5-dev] Upgrade examples, docs and tests to use angular_router 2.0


chalin:
Update:

- [x] Cheatsheet - #1076 (partially)
- Toh-5
  - [x] Prose - #1076
  - [x] Tests
- Toh-6
  - [x] Prose - #1076
  - [x] Tests (updated toh-5 tests still need to be copied over to toh-6)
- Router guide ...
  - [x] Example (update to demonstrate most 2.0 features)
  - [x] Prose (update to reflect new example)
  - [x] Tests
- [x] Ensure that all component and e2e tests of router apps are (re-)enabled
- Remove temporary
  - [x] Redirect from `firebase.json`
  - [x] Skip pattern from `scripts/config/linkcheck-skip-list.txt`
- [x] Drop or fix the currently skipped app/top-level router component tests (toh-5, -6?).

chalin:
Pushed to 5-dev.

PR #1419, opened 2018-03-22
[5-dev] toh-5: use RESTful routes, and app_route_paths.dart file

chalin:
Contributes to #1058, #1369

- Reworking toh-5 so that it is a better base for the router guide.
- All toh:
  - Switched to using ClassProvider()
  - Renamed hero_detail* to hero*


chalin:
Travis is green except for a trivial code excerpt issue. Merging.

PR #1442, opened 2018-03-28
[5-dev] Enable use of angular_analyzer_plugin + minor tutorial refactorings

### LINKED ISSUE #1436: [5-dev] Examples: enable use of angular_analyzer_plugin
Blocked on https://github.com/dart-lang/angular_analyzer_plugin/issues/494

Update the example `analysis_options.yaml` files to enable the angular_analyzer_plugin by default.

chalin:
Fixes #1436

- This enables the https://github.com/dart-lang/angular_analyzer_plugin for all examples.
- Toh-1 and toh-2 examples: "snippet" files were eliminated in favor of real Dart/HTML files that the analyzer can provide feedback on.
- Minor tweaks to app templates (removing unnecessary qualifiers from headings -- like "My" and "details!". This smooths the integration of the app into the router guide.

Staged:
- https://angulardart-org-dev.firebaseapp.com/angular/tutorial/toh-pt1
- https://angulardart-org-dev.firebaseapp.com/angular/tutorial/toh-pt2
- etc


examples/ng/doc/quickstart/analysis_options.yaml
```diff
@@ -1,7 +1,9 @@
 analyzer:
-  strong-mode: true
```
kwalrath: Why remove strong-mode? I thought that was still necessary?

chalin: As an example, if I run the analyzer from the command line, using `--strong` causes problems but `--preview-dart-2` does was we need:

```terminal
toh-3 > dartanalyzer --strong test
Analyzing test...
  error • Const variables must be initialized with a constant value at test/app_test.dart:47:22 • const_initialized_with_non_constant_value
  error • Const variables must be initialized with a constant value at test/hero_detail_test.dart:13:20 • const_initialized_with_non_constant_value
2 errors found.
toh-3 > dartanalyzer --preview-dart-2 test
Analyzing test...
No issues found!
```

Not quite related but using `--strong` with `dart` is an error:

```terminal
> dart --strong test/app_test.dart 
Snapshot not compatible with the current VM configuration: the snapshot requires 'release no-strong' but the VM has 'release strong'
```

So I figured that we should be dropping `strong-mode: true`.

In any case the code is compiled by the DDC, which is our strictest checker yet.

kwalrath: @kevmoo or @matanlurey, could you confirm what we should be using with our angular 5 examples, for now?

matanlurey: Not sure, unfortunately. I imagine keep it until it is the default?

kwalrath: Actually, @bwilkerson is probably the expert here.

bwilkerson: We're in an unfortunate place right now. I believe that DDC always has preview-dart-2 enabled, which implies strong-mode, but that dart2js doesn't enable either by default. And the command-line analyzer doesn't enable them either, but server now does.

So, assuming that users are still using dart2js for deployment, you can get the tools to be more consistent by leaving the flag for now. That said, we're hoping to flip the default everywhere fairly soon, which would mean re-visiting all of these files.

src/angular/tutorial/toh-pt1.md
```diff
@@ -83,40 +90,51 @@ Add these properties near the top of the `app_component.dart` file, just below t
...
+  <h1>{!{title}!}</h1>
+  <h2>{!{hero[!.name!]}!}</h2>
 ```
 
-Refresh the browser, and the page continues to display the hero's name.
+<i class="material-icons">open_in_browser</i>
+**Refresh the browser.** The app continues to display the hero's name.
 
-### Add multi-line template HTML
+### Show all hero properties
```
kwalrath: Can we make this name more general, rather than example specific?

chalin: Btw, these are the section headings:
screen shot 2018-03-28 at 19 06 51

So, "Add multi-line template HTML" was moved to the start.
"Show all hero properties" seems appropriate for this subsection (given that the section heading is "Show the hero").

I'm not sure how else to generalize the title.

chalin:
Ok, I'll add back strong mode.

Issue #1495, opened 2018-04-16
Cheatsheet code snippets should be excerpt from our main example sources


chalin:
Like for the test of the docs, it will make it easier to keep the code snippets up-to-date, and to detect when they aren't.

Issue #1483, opened 2018-04-12
Upgrade to Angular 5-alpha+10


chalin:
Other than the obvious upgrade in package version, changes to the examples and docs will be necessary for the following:

- [ ] `runApp()` replaces `bootstrap()` et all - #1509, etc
- [ ] Switch to static DI
- [x] Switch to using `const` Provider classes

Updates:

- [x] Examples
  - [x] pubspec package version - #1478
  - [x] architecture - #1485
  - [x] attribute-directives - #1485
  - [x] component-styles - #1485
  - [x] dependency-injection - #1491
  - [x] displaying-data - #1504
  - [x] forms - #1504
  - [x] hierarchical-dependency-injection - #1507
  - [x] lifecycle-hooks - #1504
  - [x] lottery - #1504
  - [x] pipes - #1504
  - [x] quickstart - #1481
  - [x] router - #1499. Then close https://github.com/dart-lang/angular/issues/806.
  - [x] security - #1504
  - [x] server-communication - #1504
  - [x] structural-directives - #1504
  - [x] template-syntax - #1504
  - [x] toh-0 to toh-6 - #1481, #1482 
  - [x] user-input - #1504
- [ ] Prose updates. Revisit discussions around bootstrapping:
  - [x] DI page - #1491 (also see #1490 for followup work).
  - [x] Cheatsheet - #1506
  - [ ] glossary, ... . Also related: #1495 - #1509
  - [ ] Drop link-skip rule used to exclude links to the bootstrap method API - #1509
- [x] Changelog - #1503

PR #1506, opened 2018-04-17
Cheatsheet bootstrap -> runApp() + static DI

chalin:
- This PR replaces the _Bootstrapping_ cheatsheet subtable with a _Run app and compile-time injector generation_ subtable.

  The new code excerpts need to be larger (e.g., because of the template imports, without which I doubt a single-line cheatsheet `runApp()` entry would make sense).

- All code snippets are excerpt from the quickstart app.

- The new subtable illustrates how I think the rest of the cheatsheet should be restyled:
  - Code excerpts, including the import statement in the table header, in the right column
  - Commentary on the left.

  The code column is given more space (65%)

Contributes to #1483, #1495

Staged at https://angulardart-org-dev2.firebaseapp.com/angular/cheatsheet

---

New cheatsheet entries:

screen shot 2018-04-17 at 18 01 30

Old cheatsheet entries:

screen shot 2018-04-17 at 18 02 26


kwalrath:
Interesting! Making the excerpts longer seems fine to me, and it'd make me very happy to have tested excerpts. I also like having the highlighted code. (Didn't we have that before?)

Some questions/comments:

1. Why switch the code to the right? Having the `import` statement for the whole group be above the other code seems confusing (especially in the bootstrapping code, where you have additional imports). I rather like having all the code on the left, since that seems like the most important part of the page.

2. Why have such a long name for what was formerly Bootstrapping? Do you think it'd be confusing to still call it **Bootstrapping**? (People might search for that, and the [angular.io cheatsheet](https://angular.io/guide/cheatsheet) still uses that term.)

3. What can we do to make the table take up the whole width of the page? There's a lot of wasted space on the right when the window's wide.

4. I don't like having bold font in the descriptions. It feels like bold font should be reserved for the table heads, so it's easy to scan for the next table.

chalin:
Here are some answers. I'll tackle changes tomorrow.

> 1. Why switch the code to the right? ...

It just looked strange to have a wider code column on the left. I don't mind putting it back on the right.

> 2. Why have such a long name for what was formerly Bootstrapping?

I wasn't sure about that one. I'm ok with keeping "bootstrapping". It's really about "App launching", but that seemed like a strange title to me.

> 3. What can we do to make the table take up the whole width of the page? There's a lot of wasted space on the right when the window's wide.

I think that there is a global maximum width because it can be difficult to read very wide tables. I can adjust that.

> 4. I don't like having bold font in the descriptions. It feels like bold font should be reserved for the table heads, so it's easy to scan for the next table.

Ok.

chalin:
> Having the `import` statement for the whole group be above the other code seems confusing

It isn't always an import. Regardless, again it seemed strange to have that import inside the narrow column.

Based on my comments above, let me know if your preferences have changed

chalin:
Done:

- [x] Show code in the left column
- [x] Revert to Bootstrapping title
- [x] Drop max table width
- [x] Drop bold font in the descriptions

Staging server refreshed: https://angulardart-org-dev2.firebaseapp.com/angular/cheatsheet

chalin:
Great! I'll convert the other cheatsheet tables over time (starting with the router table).

PR #1541, opened 2018-04-26
Examples: bump up min version for builder packages, such as build_runner ^0.8.6

chalin:


chalin:
Travis is failing because of broken install page links. I'll fix those in a followup PR. /cc @kwalrath

Issue #1533, opened 2018-04-24
Make the gallery the primary API reference doc for angular_components


kwalrath:
Related:

* https://github.com/dart-lang/angular_components/issues/217#issuecomment-384026372
* https://github.com/dart-lang/angular_components/issues/242

chalin:
I'm not sure that I understand this request. The gallery doesn't contain any API reference entries. In fact, in the landing page **Resources >> API reference** points to https://webdev.dartlang.org/components/api.

kwalrath:
The gallery contains HTML-oriented reference docs that are generated from the class descriptions. E.g. if you go to https://dart-lang.github.io/angular_components_example/#/app_layout, there's reference information from class-level doc comments.

chalin:
Right, but suppose you come across a `<glyph>` in an ACX app's HTML: if you [type glyph in the search box](https://dart-lang.github.io/angular_components_example/), nothing is shown. What then should the reader do?

I believe that this issue has a few parts, the main one being:

- Update references to ACX API so that readers are referred to https://dart-lang.github.io/angular_components_example

I think that to have a fuller UX we would need to enhance the ACX example app so that, for example:

- Every page provides a link back to the full API.
- Maybe the search could cover the full API.

Finally:

- We need to decide where the full API docs will be hosted from.

kwalrath:
@nshahan & @TedSander, do you have feedback on Patrice's comment?

TedSander:
So I think those are all good points, we aren't saying that the gallery is currently perfect in terms of documentation, but we think it will be a better landing page for angular components than the API documentation.

There are a couple of reasons:
a) The Dart API is not the api that we want used for interaction of a component. Many things are made public for angular templates, but they aren't for general consumption. the @Input/@Outputs/@Attributes ng-content are.
b) The examples and demos are very useful for knowing what to do with components. Internally we don't even have DartDoc, but the users find the gallery invaluable. It is consistently one of things they say is best about the widgets.
c) Mixins/Readmes provide a lot of the documentation for the components and I don't see these going into the API doc, but they are on the roadmap for the gallery.

As for the search box I like the idea of adding the selector. Glyph does not show up today because it is deprecated, but class names are already searchable (MaterialIconComponent for instance.) Please file a feature request and assign it to Nick for this.

It would also be good to fill in the holes around documentation that isn't in the gallery. Which should be relatively easy to do.

[...]

PR #1559, opened 2018-04-30
Examples: deploy ACX gallery under webdev site

chalin:
This fetches the built version of the gallery from https://github.com/dart-lang/angular_components_example/tree/gh-pages.

Staged at https://angulardart-org-dev.firebaseapp.com/examples/component-gallery

Contributes to  #1533 

cc @kwalrath @nshahan

### DIFF
--- a/gulp/example-add-apps.js
+++ b/gulp/example-add-apps.js
@@ -19,6 +19,7 @@ module.exports = function (gulp, plugins, config) {
   const rename = plugins.rename;
   const replace = plugins.replace;
 
+  const component_gallery = 'component-gallery';
   const chooseRegEx = argv.filter || '.';
   const skipRegEx = argv.skip || null;
 
@@ -28,6 +29,7 @@ module.exports = function (gulp, plugins, config) {
   const findCmd = `find ${EXAMPLES_ROOT} -type f -name ".docsync.json"`;
   const examplesFullPath = (cp.execSync(findCmd) + '').split(/\s+/).filter(p => p).map(p => path.dirname(p));
   const examples = examplesFullPath.map(p => path.basename(p))
+    .concat(component_gallery)
     .filter(p => !p.match(skipRegEx))
     .filter(p => p.match(chooseRegEx))
     .sort();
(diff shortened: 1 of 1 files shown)


kwalrath:
I like it! If @nshahan and @tedsander are good with it, I say we ship it.

We should add an index page or redirect (could be in a separate PR) so that https://angulardart-org-dev.firebaseapp.com/examples isn't a 404.

nshahan:
@chalin What would the deploy process be like on every update? 

@kevmoo what would this mean for the analytics?

kevmoo:
> @kevmoo what would this mean for the analytics?

We should update the analytics bits in the samples to match the webdev site.

Wonder if @chalin can do magic w/ a string-replace here...

chalin:
Sure: let me know what needs to be replaced, and what the replacement is.

chalin:
@nshahan - btw, if we host the ACX gallery on webdev, the app will be able to switch to using the **path location strategy**.

So who's thumbs-up are we waiting on?

chalin:
> I don't understand the context here. What does hosting the gallery on webdev solve?

A flip question is: what do we have to gain by hosting on **GH-pages** where:

- You are forced to use the dev-mode HashLocationStrategy
- You can't setup redirect rules when the API changes, etc.

Firstly, understand that there is **no extra effort required on your part**. You can continue to build and deploy the ACX gallery to the current repo.

Hosting on webdev has a number of advantages, the main ones being the following:

- UX: **webdev.dartlang.org** becomes the **central point** for all things _Angular / Dart for the Web_.

  AngularTS's main site is https://angular.io and you'll find the material design app at https://material.angular.io.

- The webdev site is versioned:

  - Stable is at https://webdev.dartlang.org and, 
  - Dev version is at https://webdev-dartlang-org-dev.firebaseapp.com
  - (We have an archive version for v3, etc)

  As has been mentioned elsewhere, we also **need the ACX gallery to be versioned**. Hosting it on webdev (and its companion stable/dev sites) solves this problem in part.

- You can use PathLocationStrategy, encode rewrite/redirect rules, ... .

(If you were asking about the original context for this PR, it is #1533.)

TedSander:
I do not want to change to PathLocationStrategy. It makes internal deployment harder, and it makes development A LOT harder.

kwalrath:
I'm going to close this, since it's been around for ages with no progress. @chalin, we can talk about this tomorrow, if that works for you.

PR #1574, opened 2018-05-04
Examples index page

### LINKED ISSUE #1573: Add an index page for /examples
From https://github.com/dart-lang/site-webdev/pull/1559#issuecomment-385464141:

> We should add an index page or redirect (could be in a separate PR) so that https://angulardart-org-dev.firebaseapp.com/examples isn't a 404.

chalin:
Adds:

- Examples landing page
- Examples entry to the sidenav

While the new examples landing page could present more information (possibly in a table format), as a first cut, this seems sufficient.

Staged at https://angulardart-org-dev2.firebaseapp.com/examples

The new Examples landing page look like this:

> screen shot 2018-05-04 at 08 08 46

Fixes #1573


src/examples.md
```diff
@@ -0,0 +1,42 @@
...
+- [QuickStart](/examples/quickstart)
+- [Tutorial](/angular/tutorial):
+  0. [The Starter App](/examples/toh-0)
+  1. [The Hero Editor](/examples/toh-1)
+  2. [Master/Detail](/examples/toh-2)
+  3. [Multiple Components](/examples/toh-3)
+  4. [Services](/examples/toh-4)
+  5. [Routing](/examples/toh-5)
+  6. [HTTP](/examples/toh-6)
+  {: start="0"}
```
kwalrath: What does this do? (Just curious)

chalin: It sets the start attribute in the `<ol>` element so that numbered items start at 0: `<ol start="0">...</ol>`.

The `0.`, `1.`, etc in the markdown are placebos :).

kwalrath: I guess I knew that part... But I didn't notice any other numbered list in the page.

chalin: Maybe what is confusing is that the Kramdown requires that the attribute be placed _after_ the targeted element. Otherwise it'll attempt to only add that attribute to the first list item.

PR #1579, opened 2018-05-08
Document webdev tool, add build_runner test cmd description

### LINKED ISSUE #1460: Link to webdev tool in build_runner page
People should be able to find info about webdev, but we shouldn't feature it yet.

/cc @kevmoo

### LINKED ISSUE #1537: Document use of webdev tool
From https://github.com/dart-lang/angular/issues/1274.

@kwalrath @kevmoo: I'm assuming that we'll still want to keep the [build_runner tool page](https://webdev-dartlang-org-dev.firebaseapp.com/tools/build_runner), but promote the use of webdev?

- [x] Update site install scripts with: `pub global activate webdev`
- [x] Docs
  - [x] Add webdev tools page
  - [x] Update Setup page
  - [x] Add page for webdev under tools
  - ...
- [x] E2E: switch to using `webdev`

Related:
- Switch all docs and uses of build_runner to webdev https://github.com/dart-lang/angular/issues/1274

chalin:
- Added webdev tool page.
- Added test command description to build_runner page.
- Across all pages, we now recommend use of the webdev CLI to build and serve apps, rather than build_runner. Note that we currently still refer to build_runner for testing.

Fixes #1537
Fixes #1460

Staged at https://angulardart-org-dev2.firebaseapp.com. Several pages are impacted, including:
- https://angulardart-org-dev2.firebaseapp.com/tools/webdev
- https://angulardart-org-dev2.firebaseapp.com/tools/build_runner
- https://angulardart-org-dev2.firebaseapp.com/dart-2
- ...

cc @kevmoo


src/tools/webdev.md
```diff
@@ -0,0 +1,56 @@
+---
+layout: angular
+title: Webdev
+description: A command line interface for Dart web development.
+---
+
+The [webdev][] package provides a command line interface (CLI) for users and
+tools to build and serve web apps. Use `webdev` instead of the deprecated `pub
+build` and `pub serve` commands. Use [build_runner][] to run tests.
```
kwalrath: build_runner came out of left field. Also, pub build and pub serve are more than deprecated, they're gone from the latest SDK. Maybe banish this stuff to a note, something like this:

...and serve web apps.

> **Related tools:** The `webdev` tool, which replaces `pub build` and `pub serve`, is built on [build_runner][]. If you've previously used `build_runner`, you should no longer need to use it directly, except to run tests.

src/tools/webdev.md
```diff
@@ -0,0 +1,56 @@
...
+
+## Installation
+
+Globally install webdev using pub:
```
kwalrath: We should probably link to pub global activate instructions. (It can be a bit tricky the first time, since you need to set the path.)

src/tools/webdev.md
```diff
@@ -0,0 +1,56 @@
...
+## Command: serve {#serve}
+
+To launch a _development server_, run the following command from your app's root
+folder:
+
+```terminal
+$ webdev serve
+```
+
+By default your app will be served from the `web` directory on port `8080`.
```
kwalrath: How do you change the default?

(It's OK to add this info later if it's not obvious, but maybe just drop the "By default" if you do.)

src/tools/build_runner.md
```diff
@@ -5,8 +5,9 @@ description: A tool for building, serving, and testing web apps (and more).
 ---
 <?code-excerpt path-base="examples/ng/doc"?>
 
-The [build_runner][] package provides a way to build, serve, and test web apps.
-Use `build_runner` instead of the deprecated `pub build` and `pub serve` commands.
+The [webdev][] tool that you normally use to build and serve your apps uses the
```
kwalrath: The first sentence is a bit hard to parse. How about:

This page describes the [build_runner][] package and its commands.

You don't usually need to use build_runner directly, except for testing. Instead, use the [webdev][] tool to build and serve apps.

src/tools/build_runner.md
```diff
@@ -31,7 +32,7 @@ $ pub get
 
 ## Command: serve {#serve}
```
kwalrath: @kevmoo should we still cover the build and serve commands? If so, I think we should move them lower in the file. Maybe:

Using a build config file
Testing with build_runner
Other build_runner commands

Should we eventually have a page dedicated just to build config files, so that people don't have to wade through the rest of the build_runner docs to be able to customize their webdev build?

src/tools/dart2js.md
```diff
@@ -7,7 +7,7 @@ permalink: /tools/dart2js
 
 Use the _dart2js_ tool to compile Dart code to deployable JavaScript.
 Another Dart-to-JavaScript compiler, [dartdevc][], is for development use only.
-The [build_runner][] tool uses dartdevc by default, but when you're compiling
+The [webdev][] and [build_runner][] tools uses dartdevc by default, but when you're compiling
```
kwalrath: uses -> use

kevmoo: webdev actually changes the default. Release is on by default for build.

Good to double-check with `webdev build --help`

src/tools/webdev.md
chalin: Excellent points and revised text, thanks.

src/tools/webdev.md
chalin: I've added some text describing how to change the default.

src/tools/build_runner.md
chalin: I've reorganized the sections as you have suggested (though with different titles). PTAL

src/tools/dart2js.md
chalin: Thanks. I've followed up with https://github.com/dart-lang/webdev/issues/50.

chalin: Ok, the combinations are getting a little too complicated.

@kwalrath, I suggest we drop (from the dart2js page) any mention of what the tooling defaults are. We'll cover tooling defaults in the respective tooling pages. WDYT?

chalin:
Updates done, staging server refreshed. PTAL

src/tools/webdev.md
```diff
@@ -0,0 +1,84 @@
...
+```
+
+To run tests directly from the command line use [build_runner test][].
+
+## Command: build {#build}
+
+Use the following command to build your app:
+
+```
+webdev build [--release] [--output <dirname>] ...
```
kevmoo: Get rid of the `[--release]` bit here and below. In general, folks will want build to do dart2js

src/tools/webdev.md
```diff
@@ -0,0 +1,84 @@
...
+
+The first build is the slowest. After that, assets are cached on disk and
+incremental builds are faster.
+
+The `build` command uses the [dartdevc][] web compiler by default. To build a
+production version of your app, add `--release`, which uses the [dart2js][]
+compiler:
+
+```terminal
+$ webdev build --release
```
kevmoo: Should emphasize a directory to output here.

`webdev build -o web:build`

src/tools/dart2js.md
kwalrath: Agreed that discussions of defaults should be in the specific tools' pages, not in the dart2js page.

src/tools/webdev.md
```diff
@@ -5,28 +5,49 @@ description: A command line interface for Dart web development.
 ---
 
 The [webdev][] package provides a command line interface (CLI) for users and
-tools to build and serve web apps. Use `webdev` instead of the deprecated `pub
-build` and `pub serve` commands. Use [build_runner][] to run tests.
+tools to build and serve web apps.
+
+<aside class="alert alert-info" markdown="1">
```
kwalrath: Sorry to change my mind, but seeing this rendered, I'm hating the big blue note here. Let's make it a regular paragraph instead, which will require some slight rewording:

...serve web apps.

The `webdev` tool is built on [build_runner][] and replaces `pub build` and `pub serve`. If you've previously used `build_runner`, you should no longer need to use it directly, except to run tests from the command line.

src/tools/webdev.md
```diff
@@ -5,28 +5,49 @@ description: A command line interface for Dart web development.
...
 $ pub global activate webdev
 ```
 
+[Globally install]: {{site.dartlang}}/tools/pub/cmd/pub-global
+
 ## Command: serve {#serve}
 
-To launch a _development server_, run the following command from your app's root
-folder:
+Use the following command to launch a _development_ server over your app
```
kwalrath: Reword slightly, to make it easier to parse:

To launch a _development server_, which serves your app and watches for source code changes, use the following command:

src/tools/webdev.md
```diff
@@ -5,28 +5,49 @@ description: A command line interface for Dart web development.
...
+and have the server watch for app source file changes:
+
+```
+webdev serve [<directory>[:<port>]]...
```
kwalrath: What does the `...` mean? I'd either be more specific (e.g. add `[<directory>[:<port>]] ...`) or drop it.

src/tools/webdev.md
```diff
@@ -5,28 +5,49 @@ description: A command line interface for Dart web development.
...
+```
+
+By default your app will be served from the `web` directory on port `8080`:
```
kwalrath: will be -> is

src/tools/webdev.md
```diff
@@ -5,28 +5,49 @@ description: A command line interface for Dart web development.
...
 
 ```terminal
 $ webdev serve
 ```
 
-By default your app will be served from the `web` directory on port `8080`.
-While the `serve` command runs, every change you save triggers a rebuild.
+You can specify different directory-port configurations. For example, visit
+[localhost:8081](http://localhost:8081) to run component tests in your browser
```
kwalrath: I found this sentence a bit hard to understand. How about:

For example, the following command sets up an additional port to run component tests in the browser at localhost:8081:

src/tools/webdev.md
```diff
@@ -5,28 +5,49 @@ description: A command line interface for Dart web development.
...
+after launching the server with these arguments:
+
+```terminal
+$ webdev serve web:8080 test:8081
```
kwalrath: Is `:8080` really necessary? (I thought it would be used by default.)

src/tools/webdev.md
```diff
@@ -48,9 +69,16 @@ $ webdev build --release
 ```
 
 You can customize your build using build config files. For more information, see
-[build_runner.][build_runner]
+[Build config files.](/tools/build_runner#config)
+
+## More information
+
+
+Run webdev with the `--help` flag or see the pub webdev package [README][webdev] for a complete
```
kwalrath: Google style is to [list the condition first](https://developers.google.com/style/clause-order), so:

For a complete list of `webdev` options, run `webdev --help` or see the [webdev package README][].

src/tools/build_runner.md
```diff
@@ -32,7 +34,7 @@ $ pub get
...
 
-To run a _development server_, use the `serve` command:
+To run a _development_ server, use the `serve` command:
```
kwalrath: Do we really need the italics here? It makes me think we're contrasting with another type of server (which we aren't doing, at least not explicitly) or defining a term (which we aren't doing here).

src/tools/dart2js.md
```diff
@@ -7,8 +7,6 @@ permalink: /tools/dart2js
 
 Use the _dart2js_ tool to compile Dart code to deployable JavaScript.
 Another Dart-to-JavaScript compiler, [dartdevc][], is for development use only.
```
kwalrath: I think we need some sort of note about you not usually using dart2js directly — usually using a tool such as webdev, with dart2js options specified in a build config file.

src/tools/webdev.md
chalin: Right (sorry that was left over from the misunderstanding I had about the `webdev build` default). I've dropped the mention of any release-mode flags here, but I included mention of the use of `--no-release` below (so that users know how to turn on the dartdevc).

chalin: Actually, with the current defaults, I think it makes more sense to present this as:

`webdev build [--no-release] --output <dirname>`

where we show `--output` as non-optional. WDYT?

src/tools/webdev.md
chalin: "..." means zero or more of the `[<directory>[:<port>]]`.

src/tools/webdev.md
chalin: It is not, thanks. I've dropped it.

src/tools/dart2js.md
chalin: In that case I'll re-introduce a form of the sentence that we had earlier, but only mention webdev :).

chalin:
Edits done and staging server refreshed. PTAL.

src/tools/webdev.md
kwalrath: Oh, funny, I hadn't noticed `...` used that way unless you show at least 2 of the thing that's being repeated. `*` would be more natural to me, except I've never seen that for command-line arguments.

src/tools/webdev.md
```diff
@@ -77,9 +76,8 @@ You can customize your build using build config files. For more information, see
 
 ## More information
 
-
-Run webdev with the `--help` flag or see the pub webdev package [README][webdev] for a complete
-list of command line options supported by the `webdev`.
+For a complete list of `webdev` options, run `webdev --help` or see the [webdev
+package README][].
```
kwalrath: Oops, this link doesn't work. (I implicitly changed its name. :)

src/tools/webdev.md
chalin: I think that it is fairly common in Linux man pages, for example see http://man7.org/linux/man-pages/man1/ls.1.html.

src/tools/webdev.md
chalin: Right! Thanks for catching that. (I had made a mental note ... but then forgot!) Fixed.

src/tools/webdev.md
kwalrath: I guess I've used BSD man pages more than Linux. They use something similar, but with no space and more obviously grouped, that makes more sense to me:

```
ls [-ABCFGHLOPRSTUW@abcdefghiklmnopqrstuwx1] [file ...]
```

chalin: Shall we use that rather than the repeated optional argument?

chalin:
@kevmoo - friendly ping.

PR #1615, opened 2018-05-24
chore: misc fixes to test scripts and example sources

chalin:


examples/ng/doc/toh-1/test/app_po.dart
```diff
@@ -1,6 +1,4 @@
-// #docregion
-
-// #docregion imports
+// #docregion , imports
```
kwalrath: This comma and space seems like something we might easily mess up. Is it better to do it this way than to have a separate `// #docregion` line? (I can see that omitting the blank line is a good thing.)

chalin: Right, the main point was to eliminate the blank line.

I'll be improving region naming with the new extractor so that we'll probably rewrite the above as follows:

    // #docregion '', imports

This will make it clearer that the default region (with the empty name) is part of the region name list. WDYT?

PR #1618, opened 2018-05-31
Removed OnActivate that should not be there, yet

gazialankus:
The class does not implement OnActivate at this point. It will be added a bit below.


gazialankus:
I see, these excerpts are connected to the code in the repository, nice. 
I tried to do it, did it work? 
Thank you so much.

chalin:
Yes, perfect! Thanks for the fix!

PR #1629, opened 2018-06-05
Examples: upgrade SDK (2-dev.60) and Angular et all (+14) w/o ACX

chalin:
Contributes to #1628


chalin:
Note: this PR
- Does not set `DART_VM_OPTIONS`
- Does not include ACX +14 (since it hasn't yet been released)

PR #1633, opened 2018-06-08
Examples: doc region cleanup

chalin:
- Use quoted form (`''`) of default region name. The unquoted form is deprecated.
- Remove unnecessary initial `#docregion` tags since, by default, the new code_excerpter assumes that the entire file corresponds to the default region.
- A few other misc code region fixes.


examples/ng/doc/attribute-directives/lib/app_component.html
```diff
@@ -1,4 +1,4 @@
-<!-- #docregion v2, -->
+<!-- #docregion '', v2 -->
```
kwalrath: Why is the `''` necessary here? I assumed that every file has a `#docregion` starting at the beginning, unless declared otherwise. Is it having the named docregion is specified here, at the beginning, that somehow makes the `''` be required?

chalin: The region named `''` will, by default, refer to the entire file. When it does, then there is no need to "name" this default region.

If you want the `''` region to refer to something less than the entire file (as is the case in this here), then you need to explicitly mark a start (line 1) and ending (line 28 for this file) for the `''` region.

PR #1661, opened 2018-06-25
Fix misspelling

JacobAMason:
natigated -> navigated

**IMPORTANT!** Submit your PR against the right branch:
- `4.x` is for AngularDart 4.0 stable (https://webdev.dartlang.org)
- `master` is for AngularDart 5.0 preview (https://webdev-dartlang-org-dev.firebaseapp.com)


JacobAMason:
I won't be signing the CLA, but this is a simple misspelling, so I'm sure someone can throw it in with another change to this file.

chalin:
Thanks! Fixed.

PR #1666, opened 2018-06-27
Updated fetch-data JSON tutorial

### LINKED ISSUE #1663: Upgrade tutorials/get-data/fetch-data to Dart 2
Context: https://github.com/dart-lang/site-webdev/issues/1634#issuecomment-400126200

chalin:
Fixes #1663

(For now we only build fetch_data. Tests will be added later.)

Staged at https://angulardart-org-dev2.firebaseapp.com/tutorials/get-data/fetch-data


src/_tutorials/get-data/fetch-data.md
```diff
@@ -169,67 +164,70 @@ void showJson(Event e) {
     'favoriteThings': favoriteThings
   };
 
-  // Convert everything to JSON and display the results.
```
kwalrath: I still see the old comment in the DartPad.

src/_tutorials/get-data/fetch-data.md
```diff
@@ -169,67 +164,70 @@ void showJson(Event e) {
...
 
-The dart:convert library contains two convenient functions
+The `dart:convert` library contains two convenient functions
 for working with JSON strings:
 
 | dart:convert function | Description |
 |---|---|
-| <a href="{{site.dart_api}}/{{site.data.pkg-vers.SDK.channel}}/dart-convert/JsonCodec/decode.html" target="_blank" rel="noopener">JSON.decode()</a> | Builds Dart objects from a string containing JSON data. |
-| <a href="{{site.dart_api}}/{{site.data.pkg-vers.SDK.channel}}/dart-convert/JsonCodec/encode.html" target="_blank" rel="noopener">JSON.encode()</a> |  Serializes a Dart object into a JSON string. |
+| [json.decode()][]{:target="_blank" rel="noopener"} | Builds Dart objects from a string containing JSON data. |
```
kwalrath: Should we be using jsonDecode and jsonEncode instead? (That's what we're showing in the library tour.)

src/_tutorials/get-data/fetch-data.md
```diff
@@ -239,78 +237,79 @@ void showJson(Event e) {
...
-its items are comma-separated.
-The list in this example contains strings.
-A map is delineated with curly brackets;
-it contains comma-separated key/value pairs,
-where the key appears first, followed by a colon,
-followed by the value.
-In this example,
-the keys in the map are strings.
-The values in the map vary in type but they are all JSON-parsable.
+- **Numeric** and boolean values
```
kwalrath: I like this formatting change!

src/_tutorials/get-data/fetch-data.md
```diff
@@ -402,74 +400,86 @@ the app makes a GET request of the server
...
-A Future is a way to perform potentially time-consuming operations,
+The `getString()` method uses a Future object to handle the request.
+A [Future][] is a way to perform potentially time-consuming operations,
 such as HTTP requests, asynchronously.
 If you haven't encountered Futures yet,
 you can learn more about them in
 [Asynchronous Programming: Futures]({{site.dartlang}}/tutorials/language/futures).
-Until then, you can use the code above as an idiom
-and provide your own code for the body of the processString() function
+Until then, you can use the code above as an guide
```
kwalrath: an -> a

src/_tutorials/get-data/fetch-data.md
```diff
@@ -509,47 +520,55 @@ Because the request in this example is a simple GET request,
...
+      wordList.children.add(li);
+  }
+}
+{% endprettify %}
+
+If the status code is 200,
+the file was found and loaded successfully.
+The content of the requested file (`portmanteaux.json`) is
+returned in the [responseText][] property of an HttpRequest object.
+For any other status code, an error is reported.
```
kwalrath: Let's reword the last sentence to not be passive voice. (This paragraph uses a lot of passive voice when the "doer" isn't the code you see. ) Or maybe we can just drop the last sentence, since we don't discuss (here) what happens on success? Either way, let's use active voice if we're describing the code that they're actually seeing.

I'm sure you thought of this, but it seems asymmetric to show the LIElement code in the error handler but not in the success handler. Would it be too onerous to put the "default:" code into its own method, maybe `reportFailure(request.status)`? That way, we avoid having to either describe it or gloss over it. It wasn't an issue before because the 200: code was inline and explained here.

src/_tutorials/get-data/fetch-data.md
```diff
@@ -558,21 +577,43 @@ contains a JSON-formatted list of strings.
...
 and sends it as a single string
 to the client program.
-The client program receives the JSON string
-and uses JSON.decode()
-to create the String objects specified by the JSON string.
 
-<img class="scale-img-max" src="../images/json-parse.png"
-     alt="Decode a JSON formatted list of strings">
+Using [json.decode()][],
+the code easily converts the JSON-formatted list of words
```
kwalrath: the code -> the client
?

src/_tutorials/get-data/fetch-data.md
chalin: AFAIK, `jsonDecode` and `jsonEncode` are convenience methods ("aliases") used to help folks transition Dart 1 code to Dart 2 when there is a `json` variable declared in scope. 

We don't have a `json` variable declared in scope so I'd prefer using the base methods.

src/_tutorials/get-data/fetch-data.md
chalin: I changed it to "the app".

src/_tutorials/get-data/fetch-data.md
chalin: > Or maybe we can just drop the last sentence

Done

> ... but it seems asymmetric to show the LIElement code in the error handler but not in the success handler

Agreed. I've elided the error handling code. (We've done the same at the start of the tutorial in the code excerpt showing `makeRequest()`.)

chalin:
Updates done and staging server refreshed. PTAL

src/_tutorials/get-data/fetch-data.md
kwalrath: It's just a little gross to use the library prefix as if it were real API. @kevmoo, have an opinion?

chalin:
@kwalrath - FYI: while walking through the debugging notes, I was surprised by one of the runtime types, so I turned on no-implicit-dynamic, and then tweaked the code a little.

src/_tutorials/get-data/fetch-data.md
```diff
@@ -134,16 +134,14 @@ void _populateFromJson() {
   favTwo.value = jsonData['favoriteThings'][1];
   favThree.value = jsonData['favoriteThings'][2];
 
-  if (jsonData['chocolate']) {
-    loveChocolate.checked = true;
-  } else {
-    noLoveForChocolate.checked = true;
-  }
+  final chocolateRadioButton =
```
kwalrath: No need to change, but just fyi...

I had a harder time understanding this code than the previous code. I think I saw the `false` and assumed chocolateRadioButton was a boolean (even though the type is right there in the name). And `?:` is always a little harder to scan (for me, at least) than if-else. Is there a performance reason for not using `if (jsonData['chocolate'])`?

src/_tutorials/get-data/fetch-data.md
kwalrath: OK by me if it's OK by @kevmoo.

src/_tutorials/get-data/fetch-data.md
chalin: The main reason is structural: the new code makes it clear that exactly one of the radio buttons in the group is being set to "checked". I.e., the first line chooses the radio button, and the second line sets it.

src/_tutorials/get-data/fetch-data.md
chalin: @kwalrath - in a [similar situation](https://github.com/dart-lang/stagehand/pull/471#discussion_r178179616), @kevmoo wrote:

> Either [`json.decode` or `jsonDecode`] is fine, honestly.

Shall I wait or go ahead and merge?

kwalrath: Merge!

PR #1670, opened 2018-06-29
Upgrade to SDK 2-dev.66

chalin:


chalin:
Travis is red for the component tests but only because the archiving (caching) step takes too long. I'll merge this for now and split the tests into two tasks in a followup PR.

PR #1714, opened 2018-07-18
Dependency injection page cleanup

chalin:
Staged at https://angulardart-org-dev.firebaseapp.com/angular/guide/dependency-injection


examples/ng/doc/dependency-injection/lib/app_component.dart
```diff
@@ -30,7 +28,6 @@ import 'src/providers_component.dart';
     CarComponent,
     HeroesComponent,
     InjectorComponent,
-    TestComponent,
```
chalin: This page had a component used to illustrate testing, but this was a leftover from the Angular TS days. We don't even want to suggest testing in that way so I've dropped this component from the example app.

PR #1726, opened 2018-07-21
Examples: upgrade to ACX beta+1

chalin:


chalin:
One Travis job is failing because of bad links into API docs. I'll fix that in a followup PR.

Issue #1747, opened 2018-07-27
Phase out AngularJS directives


chalin:
- [ ] live-example
- ...

PR #1748, opened 2018-07-27
Replace live-example AngularJS directive by a custom Jekyll tag

chalin:
In this way, the live example links will be generated statically, rather client-side.

Contributes to #1747 

Staged at https://angulardart-org-dev2.firebaseapp.com, for example see https://angulardart-org-dev2.firebaseapp.com/angular/guide/attribute-directives

cc @filiph


chalin:
Travis is green, this is mainly an infrastructure PR, and you're probably busy so I'll go ahead and merge.

PR #1783, opened 2018-08-14
Jekyll plugin cleanup; replace bare `<code-example>` AngularJS by markdown

chalin:
There are no effective doc changes.

- Drop Jade plugin since we won't be using Jade anymore
- General Jekyll plugin cleanup
- Replace bare occurrences of `<code-example>` AngularJS directive by markdown code fences


kwalrath:
Whoa, this isn't new, but check out the bad table formatting in https://angulardart-org-dev.firebaseapp.com/angular/guide/template-syntax#binding-targets:

image

kwalrath:
The glossary entries looked different (all blue). Is that intentional?

chalin:
Whoa indeed! I'll fix that in a followup PR.

chalin:
> The glossary entries looked different (all blue). Is that intentional?

Markdown code blocks are statically pretty printed using Rouge. I did find it strange that it was all the same color. I'll see if I can resolve this in the next PR, ok?

PR #1790, opened 2018-08-16
Add update info to the webdev page

kwalrath:
Followup to dart-lang/stagehand#574.


kwalrath:
Yeah, I thought we add it, too. Maybe we do somewhere else...

PR #1801, opened 2018-08-22
Migrate to jekyll-assets 3.x

chalin:
Fixes #1329

PR #1802, opened 2018-08-22
Assets cleanup: use conventional folder names, simplify jekyll-assets config

chalin:
Followup to #1801.

- Un pin font-awesome-sass
- Use standard folder names for asset sources. While folders have been renamed under `src`, this doesn't impact where the assets end up in the generated site.
- Jekyll: drop custom asset path config now that we're using conventional paths.


kwalrath:
> ...The only exception is the rename of `covers` images to `cover` (and this only impacts the paths in the books page).

Irrelevant for site-webdev, right?

chalin:
> Irrelevant for site-webdev, right?

Right (I've removed that copy-paste error), thanks!

PR #1813, opened 2018-08-31
added section to direct Linux users to modify the PATH variable.

oneyb:
I think I got this commit right.

Should help alleviate problems related to https://github.com/dart-lang/sdk/issues/19926


oneyb:
I signed the CLA. HTH.

kwalrath:
Thanks for the contribution! These files are tricky because they're in a shared repo. Could you try creating a PR for the https://github.com/dart-lang/site-shared repo?

(I'll file a bug about making this clearer in our docs.)

PR #1839, opened 2018-10-27
Remove build_runner page; add build_runner test to webdev page

### LINKED ISSUE #1825: 'Build_runner' page is out of date
From URL: https://webdev.dartlang.org/tools/build_runner

At a minimum we should remove the section about [build_config_files](https://webdev.dartlang.org/tools/build_runner#build-config-files), these should almost never be used any more.

We now have "release" mode as a real supported thing, and you can configure builders differently based on that mode with `release_options` and `debug_options`, which override `options` for their respective modes.

kwalrath:
It's moving to site-www (dart-lang/site-www#1212). I'm moving the web-specific content to the webdev page and renaming that page.

Fixes #1825.

Staged at URLs like:

https://kw-webdev-dartlang-1.firebaseapp.com/tools/webdev
https://kw-webdev-dartlang-1.firebaseapp.com/tools/build_runner (redirects to webdev page)


kwalrath:
Somehow, after the merge, this PR retained the index.md files that @chalin's PR had dropped. He stripped out those files in a followup commit, eliminating the link check failures.

kwalrath:
@jakemac53 & @natebosch, this is ready for review and staged at:

https://kw-webdev-dartlang-1.firebaseapp.com/tools/webdev

src/tools/webdev.md
```diff
@@ -1,30 +1,86 @@
...
-[Globally install][] webdev using pub:
+<aside class="alert alert-info" markdown="1">
+  **Dart 2 note:**
+  The `webdev` tool replaces the Dart 1.x `pub build` and `pub serve` commands.
+</aside>
+
+## Setup
+
+To use `webdev`, you first need to install it.
+The `build_runner` tool isn't installable;
```
jakemac53: This isn't entirely accurate, the main difference is that `webdev` is **globally** installed (which means it will be on your PATH if you have  your global pub cache on your PATH), while `build_runner` is only **locally** installed, and won't be on your PATH.

Maybe for the purposes of this comment it is fine as is though?

src/tools/webdev.md
```diff
@@ -86,17 +140,58 @@ top-level `web` folder into the `build` directory:
...
+```terminal
+$ pub run build_runner test -h
+```
+
+Arguments after the empty `--` argument
+are passed directly to the [test package][] runner.
+To see all command-line options for the test package runner, use this command:
+
+```terminal
+$ pub run build_runner test -- -h
```
jakemac53: Note that this will run a build - we could say just run `pub run test -h` which won't... but that might be confusing for other reasons.

src/angular/note/changelogs.md
```diff
@@ -0,0 +1,22 @@
+---
+title: "Changelogs"
```
chalin: This file shouldn't be here; it was removed via a recent PR. What _this_ PR rebased? If so, then please just manually delete the file.

src/tools/webdev.md
chalin: I agree with Jake's comments. IMHO we should **not** state that "`build_runner` tool isn't installable". We do give the steps further below concerning the addition of `build_runner ` as a dev dependency. Somehow the prose should tie into that.

src/tools/webdev.md
```diff
@@ -63,10 +119,8 @@ following command changes the test port from the default (8081) to 8083:
 $ webdev serve web test:8083 # App: 8080; tests: 8083
```
natebosch: Out of scope - but this command wouldn't serve anything on `8080` AFAIK, unless `webdev` is doing some other arg  manipulation...

src/tools/webdev.md
```diff
@@ -86,17 +140,58 @@ top-level `web` folder into the `build` directory:
 $ webdev build --no-release --output web:build
 ```
 
-You can customize your build using build config files. For more information, see
-[Build config files.](/tools/build_runner#config)
+
+### build_runner test {#test}
+
+Use the `build_runner test` command to run your app's [component tests][]:
```
natebosch: Should we state again that this requires a dep on `build_test`? Might not be necessary since the command itself will tell you if you forget...

src/angular/note/changelogs.md
kwalrath: I'm having git issues with this one. I might just start fresh with a new PR.

chalin: Anything I can help you with?

kwalrath: Possibly... I've given you details in chat. (Thanks for offering!)

src/tools/webdev.md
kwalrath: The build takes a while, so I think your suggestion is better.

src/tools/webdev.md
kwalrath: Just tested again, and indeed it does magically serve the app on 8080.

src/angular/note/changelogs.md
kwalrath: Fixed by creating a new branch and cherry picking changes from the broken branch to the new branch.

src/tools/webdev.md
kwalrath: Or maybe I should: the command's output wasn't very helpful. I saw this when I took "build_test" out of the dev dependencies for [quickstart](https://github.com/dart-lang/site-webdev/tree/master/examples/ng/doc/quickstart):

```
...
00:01 +0 -1: compiling test/app_test.dart [E]                          
  Failed to load "test/app_test.dart": Failed to load script at http://localhost:52707/ffnDetX%2Bi88zZ8BKHE3b%2BZ46HJO4IHyx/test/app_test.dart.browser_test.dart.js.
...
```

natebosch: Weird, I wonder if it was maybe still in the transitive dependencies? We actually need it to be in the immediate dependencies.

Here is the check:
https://github.com/dart-lang/build/blob/f61f0ca09de873eb141d3d1ac816832ca8deda1f/build_runner/lib/src/entrypoint/test.dart#L147-L151

natebosch: On second though, even if it isn't an immediate dep it shouldn't matter - it applies on the root package, not dependencies:

https://github.com/dart-lang/build/blob/f61f0ca09de873eb141d3d1ac816832ca8deda1f/build_test/build.yaml#L20

@jakemac53 - any idea what might be happening here? Why doesn't the `build_test` builder work if it's a transitive dep?

jakemac53: It was probably actually a missing `build_web_compilers` dependency? We don't warn about that one for the `test` command since you can run vm tests without it....

jakemac53: We could potentially try and parse the output of package:test to recognize this error and give a better one.

src/tools/webdev.md
kwalrath: I've changed the wording. ptal.

src/tools/webdev.md
kwalrath: I left build_web_compilers in there... build_test was the only thing I removed.

kwalrath:
I've updated the text: @jakemac53 & @natebosch, please take another look.

kwalrath:
Is this page good to go, once the site-www page is published?

PR #1844, opened 2018-11-13
API reference: by default, show Dart SDK entries only

chalin:
- [x] By default, show Dart SDK entries only
- [x] Hide Angular amulet and version # from page heading

Staged, e.g., see:

- https://angulardart-org-dev3.firebaseapp.com/api
  shows SDK entries only
- https://angulardart-org-dev3.firebaseapp.com/api?package=Dart%20SDK
  shows SDK entries as it did before
- https://angulardart-org-dev3.firebaseapp.com/api?package=angular
  shows Angular entries as it did before
- ...
- To see all packages, select **ALL** from the **Packages** dropdown.

cc @kevmoo


kwalrath:
Cool! I'll send you some new text for the intro.

kwalrath:
Here are the content changes I was thinking of. I made the library list code font instead of boldface because things were looking way too bold.

```
    template:
      '<p>' +
      '  This page lists API from ' +
      '  <a href="' + DART_API + '"><b>Dart SDK libraries</b></a> ' +
      '  that Dart web apps frequently use: ' +
      '  <span ng-bind-html="$ctrl.sdkLibList | trustAsHtml"></span>.' +
      '  Use the <b>PACKAGES</b> dropdown to view API from Angular libraries.' +
      '</p>' +
...
      function list2text(list) {
        var codeFontList = list.map(function(e) { return '<code>' + e + '</code>'; });
        var and = list.length > 2 ? ', and ' : ' and ';
        return codeFontList.slice(0, list.length - 1).join(', ') + and + codeFontList[list.length - 1];
      }
```

chalin:
> Here are the content changes I was thinking of.

Done. PTAL (the staging server has been refreshed as well).

> I made the library list code font instead of boldface because things were looking way too bold.

Nice change, thanks.

kwalrath:
Looking good! How do we remove the version from the heading?

kwalrath:
> Only on this page?

The Angular version number can remain on angular-specific pages, so... yes?

kwalrath:
I think there are scattered other pages across the site that have angular version #s but aren't angular-specific, but they can be handled separately.

chalin:
Done. PTAL

kwalrath:
I'll commit this once I've made sure that pages that link to the angular API reference have the right URLs.

kwalrath:
I see only one reference on this site to "/api", and I'll fix that soon in a separate PR.

Issue #1766, opened 2018-08-07
'Imperative Navigation and Route Parameters' page issue


jeffdahl:
From URL: https://webdev.dartlang.org/angular/guide/router/3

The page states, "Notice that the hero path ends with `:id`."  However, nowhere on the page is there a path that ends with `:id`.  Should this be `:$idParam`?

chalin:
Yes, good catch! I'll fix that tomorrow.

PR #1860, opened 2018-12-07
Change `:id` to `:$idParam`

artem-galas:
#1766


kwalrath:
Thanks very much for bringing this to our attention, @artem-galas! It looks like Patrice has a different fix ready, so I'll close this in favor of #1861.

PR #1863, opened 2018-12-09
Drop the "Learning Angular" page

### LINKED ISSUE #641: [ng] Drop the "Learning Angular" page
The reworked [Angular TS site](https://aio-staging.firebaseapp.com/) no longer has a "Learning Angular" page. I've always found this page to be somewhat redundant. We should consider removing it too.

cc @kwalrath

artem-galas:
Fixes #641 

The most probably I've removed all notice about `learning-angular` page but, please double check before merge.


chalin:
Hi @artem-galas. Thanks for the PR, but I don't believe that this is a change that we want to entertain at the moment. @kwalrath WDYT?

kwalrath:
@artem-galas has done some good work here. If we do indeed want to remove this page I'd be tempted to accept this PR, but we'd need even more changes (redirect, remove any images used only by this page, ...). Also, we'd need to make sure that this page is indeed not needed.

I took a quick look at analytics. This is the 15th most visited page on the site, and the 8th most popular page under /angular. It's only the 48th most popular landing page, though, so most people come to it from within the site. Average time on this (very short) page is 38 seconds, compared to an average of almost 2 minutes. The bounce rate and exit rates are relatively low, so people are finding useful links on this page. But I suspect they could find those links elsewhere.

@chalin you know the site better than I do, so I leave it to your judgment.

chalin:
Thanks for the feedback and the analytics. Ok, I'll leave this open, though I might not be able to take a good look until the new year.

kwalrath:
Patrice wasn't able to get to this. If someone else is willing to do the work of reviewing this PR and/or making all the requisite changes, we'd welcome the help. Otherwise, I'll close it.

PR #1883, opened 2018-12-25
rename `getAll()` method to `getHeroes()`

wurikiji:
According to document itself, the method `getAll()` should be `getHeroes()`.


wurikiji:
Thank chalin, I create a new PR.

PR #1886, opened 2018-12-28
Rename `getAll()` to `getHeroes()` on sample code

wurikiji:
I applied the comment on #1883

### DIFF
--- a/src/angular/guide/dependency-injection.md
+++ b/src/angular/guide/dependency-injection.md
@@ -266,7 +266,7 @@ defined in its own file.
   import 'mock_heroes.dart';
 
   class HeroService {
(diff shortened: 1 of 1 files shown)


kwalrath:
The build is failing because the following source files (under [/examples/ng/doc/dependency-injection](https://github.com/dart-lang/site-webdev/tree/master/examples/ng/doc/dependency-injection)) need to be updated to use getHeroes instead of getAll:

lib/src/heroes/hero_service_1.dart
lib/src/heroes/hero_service.dart
lib/src/injector_component.dart

chalin:
@wurikiji - will you be following up with another PR? If so, great! If not that's ok, just let us know and we can handle the changes to the prose.

PR #1913, opened 2019-02-19
Fix import bug in Tour of Heroes tutorial pt6

kaesle:
Part 6 of of the ToH tutorial creates a lib/-level
service and imports it using the name
'angular_tour_of_heroes' in the path where the current
version (running webdev 1.0.0) seems to require
'angular_app' instead.


kaesle:
CLA complete.

kwalrath:
Thanks for caring enough to create this PR! I'm going to close it because it turns out that the `angular_tour_of_heroes` name is intentional. We could make this clearer, so I'm leaving #1914 open.

PR #1924, opened 2019-05-07
Replace with redirects deployment

mit-mit:


.firebaserc
```diff
@@ -1,17 +1,5 @@
 {
   "projects": {
     "default": "webdev-dartlang-org",
```
kwalrath: Remove the comma?

mit-mit: Good catch

PR #1925, opened 2019-05-07
Update README.md

mit-mit:


kwalrath:
I don't know why the build failed, but the new links in the readme don't look valid.

PR #1935, opened 2019-06-17
Add optional .html extension to redirect source urls.

msmorgan:
Some Google results about AngularDart still have a .html extension on
the URL. This commit adds .html to each of the redirect source blobs, so
that the redirect will happen with or without the extension.


msmorgan:
I was not able to test locally whether this works. Firebase complained that I don't have permission for "firebase.projects.get".

mit-mit:
Thanks, LGTM

mit-mit:
@msmorgan this has landed, and from my sporadic testing appears to be working. Can you try with the links you had seen broken before?

msmorgan:
@mit-mit Thank you! I tried with the links that were broken before, and they're working now.

PR #1936, opened 2019-07-01
Redirect old Angular API pages to new equivalents.

msmorgan:
Use a [capture segment][1] to point redirects from old documentation
URLs to the same page on the new site.

[1]: https://firebase.google.com/docs/hosting/full-config#capture_url_segments_for_redirects


kwalrath:
Thank you, @msmorgan!

